PR #25, opened 2026-06-30
plain triage: never end the turn waiting on a background sub-agent

tella-butler:
## Problem
The Plain ticket-triage automation runs as a **single unattended turn**, but the agent could spawn a background/async sub-agent and then **end its turn intending to resume "when it reports back."** Nothing resumes an unattended run, so the run completed with **no triage note posted** — the entire investigation was lost even though the run logged `finished ok`.

Observed twice during the post-maintenance webhook backfill on 2026-06-30 (thread `th_01KWANCHM7RQH69BFX60VJVJZ1` — "zero-second videos"). Both runs reached a high-confidence diagnosis, then self-terminated waiting on a background codebase agent; the thread stayed `TODO` with zero notes. The diagnosis had to be recovered manually from the session transcript and posted by hand.

## Fix
Prompt-only change: add an explicit instruction that the run is one unattended turn, that the Agent tool returns sub-agent findings **inline**, and that the note must be posted **within the turn** — never stop to wait for a background agent.

## Note on rollout
The live triage automation stores its prompt in the automation record (`~/.backstage-automations/`); `ensureTriageAutomation` only seeds from this file when absent. So merging this updates the source-of-truth + future seeds, but the **running** automation's stored prompt must be re-synced separately to take effect live.


tella-butler:
Closing — superseded on master: fixed at the runner level instead (16cd79ef holds the turn boundary open while background sub-agents run), which covers triage and every other run kind. — Michael

PR #34, opened 2026-07-01
Collapse "Acting as" into a submenu

tella-butler:
## What

The Michael menu (behind the top-bar title) showed the full "Acting as" teammate list inline, which took up most of the dropdown. This collapses it into a single **Acting as** row that displays the current user and reveals the teammate list in a flyout submenu on hover/click.

## Details

- `SettingsMenu.tsx`: the inline user list is now a `settings-submenu` trigger row (avatar + "Acting as" + current user + chevron) that opens a `settings-submenu-panel` flyout containing the same teammate list. Submenu collapses when the menu closes or a teammate is picked.
- `global.css`: styles for the trigger row and flyout panel, plus a transparent hover bridge over the gap so moving the pointer from the trigger to the panel doesn't dismiss it.

Only my two hunks were committed; the shared checkout's other in-progress edits were left untouched.

🤖 Created by [this Michael session](https://michael.taila5d766.ts.net/backstage/session/bks-019f1e1a-f26c-7000-9a61-8aa1163d3220)

### DIFF
--- a/src/frontend/components/SettingsMenu.tsx
+++ b/src/frontend/components/SettingsMenu.tsx
@@ -14,6 +14,7 @@ export function SettingsMenu({
 	connected?: boolean;
 }) {
 	const [open, setOpen] = useState(false);
+	const [actingOpen, setActingOpen] = useState(false);
 	const ref = useRef<HTMLDivElement | null>(null);
 	const currentUser = useCurrentUser();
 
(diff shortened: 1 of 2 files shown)


tella-butler:
Closing — stale: SettingsMenu was rebuilt (Tailwind bottom-sheet with the Acting-as avatar list inline); this submenu change targets the old menu structure. — Michael

PR #35, opened 2026-07-01
Center the session name in the mobile top bar (iOS-sheet style)

tella-butler:
On mobile the session name sat in a second header row *below* the Back chevron, which felt disconnected. This moves it up next to Back and centers it across the top bar like an iOS sheet nav bar.

### What changed
- Render the session title centered in the mobile `.app-header`, next to the Back chevron. It's absolutely positioned with symmetric side insets so it stays honestly centered on the bar and clears the Back button, truncating with an ellipsis rather than pushing the layout.
- Drop the now-redundant in-header title row (`.viewer-title`) on phones, so the header below is just the compact actions row.
- Session-only: other pushed views keep their existing left-aligned `.detail-topbar-title`.

Desktop is unaffected — `.app-header` is `display:none` above 720px, so the centered title only appears on mobile.

🤖 Created by [this Michael session](https://michael.taila5d766.ts.net/backstage/session/bks-019f1efb-ef7c-7000-b8b8-b2bb1b0c4b6f)

### DIFF
--- a/src/frontend/App.tsx
+++ b/src/frontend/App.tsx
@@ -509,6 +509,12 @@ function App() {
 									? "New session"
 									: "";
 
+	// iOS-sheet style: on a mobile session page the session name sits centered in
+	// the top bar, next to the Back chevron. Other pushed views keep their plain
+	// left-aligned .detail-topbar-title, so this is session-only.
+	const mobileHeaderTitle: string =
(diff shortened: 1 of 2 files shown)


tella-butler:
Closing — superseded on master: the mobile top-bar title is now absolutely centered iOS-style with a title/subtitle column. — Michael

PR #40, opened 2026-07-01
Show in the chat how long the agent has been working

tella-butler:
The terminal spinner shows elapsed time ("9m, 7.3s") while Claude works; the Backstage chat had no equivalent — you couldn't tell how long Michael had been at it.

Now:

- The busy bar above the composer reads **"Michael is working… 9m 7s"**, ticking once a second.
- The live work block header ticks **"Working for 9m 7s"** against the wall clock, instead of only advancing when a new tool result lands. Finished blocks keep the existing static "Worked for …".

Implementation notes:

- The timer anchors on the current turn's user message timestamp (server truth, so it's correct when you open a session mid-run), falling back to the optimistic just-sent bubble, or to when the client saw the run begin (e.g. automation turns).
- The once-a-second tick lives in a small isolated `Elapsed` component, so it re-renders one span per second rather than the whole `SessionViewer`/transcript. Tabular numerals so the digits don't jitter.

🤖 Created by [this Michael session](https://michael.taila5d766.ts.net/backstage/session/bks-019f1f6c-e013-7000-af99-bce31f53023a)

### DIFF
--- a/src/frontend/components/Elapsed.tsx
+++ b/src/frontend/components/Elapsed.tsx
@@ -1,0 +1,34 @@
+import React, { useEffect, useState } from "react";
+
+/**
+ * Self-ticking "how long has this been going" label (CLI-spinner style).
+ * Isolated so the once-a-second tick re-renders only this span, not the
+ * component that embeds it.
+ */
+export function Elapsed({
+	since,
+	className,
+}: {
+	since: number;
+	className?: string;
+}) {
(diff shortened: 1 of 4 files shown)


tella-butler:
Closing — superseded on master: the chat shows a busy dot + self-ticking elapsed timer (busy-elapsed in SessionViewer, elapsedClock in Sidebar/CatchUpDeck). — Michael

PR #41, opened 2026-07-01
Repo chips in the session top bar, in front of the workspace title

tella-butler:
The repo chips (primary repo + attached repos + `+ repo` attach button + fresh-session repo switcher) used to render as a separate `RepoBar` row under the session header. Now they render inline in the session top bar, in front of the workspace title — the top bar reads **repo · workspace title** and the separate repo row is gone, so the repo isn't displayed twice.

- `SessionViewer`: `RepoBar` moved into `viewer-title` (after the source chip, before the title), standalone row removed; compact-header collapse threshold bumped 820 → 900px for the extra chip width.
- CSS: `.repobar` loses its row padding/border and becomes a non-shrinking inline flex group (the title next to it is what truncates); on phones the `+ repo` button is hidden (chips stay) so the title keeps its width.

All RepoBar behavior (attach, detach, switch primary on fresh sessions) is unchanged — only where it renders.

🤖 Created by [this Michael session](https://michael.taila5d766.ts.net/backstage/session/bks-019f1f83-80ca-7000-b72d-af417208a856)

### DIFF
--- a/src/frontend/components/SessionViewer.tsx
+++ b/src/frontend/components/SessionViewer.tsx
@@ -751,9 +751,10 @@ export function SessionViewer({
 		return () => ro.disconnect();
 	}, [topbarEl]);
 	// Collapse before the inline row can overrun: the title's non-shrinkable
-	// floor (source chip + model pill + Working pill) plus all six action buttons
-	// needs ~820px, so below that the secondary actions move into the ⋯ menu.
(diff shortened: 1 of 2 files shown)


tella-butler:
Closing — stale: the session top bar was rebuilt since; repo visibility went a different way (RepoBar in the session viewer + sidebar repo chips). Conflicting for ~3 weeks. — Michael

PR #49, opened 2026-07-03
Composer: mobile-friendly "+" add menu (bottom sheet)

tella-butler:
## Problem

On mobile, the "+" (add) button in the chat composer input opened a small desktop-style popover — cramped touch targets and an awkward floating card that isn't mobile-friendly.

## Fix

On phone widths (`useIsPhone`, ≤720px) the add menu now opens as an iOS-style `BottomSheet`, the same primitive already used for the Settings and Michael menus — full-width rows, comfortable thumb targets, backdrop/Esc/drag-to-dismiss. Desktop is unchanged (keeps the `.composer-menu` popover).

Also skips the outside-click listener while the phone sheet is open, since the sheet portal owns its own dismissal.

🤖 Created by [this Michael session](https://michael.taila5d766.ts.net/backstage/session/bks-019f2721-eada-7000-aead-b39bd99c0530)

### DIFF
--- a/src/frontend/components/Composer.tsx
+++ b/src/frontend/components/Composer.tsx
@@ -16,6 +16,7 @@ import {
   IconChevronDown,
 } from "./icons";
 import { Tooltip } from "../ui/tooltip";
+import { BottomSheet } from "../ui/sheet";
 import { VoiceInput } from "./VoiceInput";
 import { useIsPhone } from "../hooks/useIsPhone";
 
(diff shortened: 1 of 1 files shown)


tella-butler:
Closing — stale: the composer add menu was reworked since (composer-add-btn with its own mobile styling/media queries), and this branch conflicts with all of it. Redo from current master if a true bottom-sheet is still wanted on phones. — Michael

PR #51, opened 2026-07-06
Mobile: change repo from the ⋯ menu + show 'repo · model' under the title

tella-butler:
## What & why

On mobile there was no way to change a session's repo: on phones the whole header title row (which carries the inline \`RepoBar\`) is CSS-hidden — the centered top-bar title replaces it — so the repo switcher never rendered.

This adds repo switching to the phone **⋯ overflow menu**, and surfaces the current repo in the subtitle under the title as **\`repo · model\`**.

## Changes

- **New \`RepoSwitchMenu\`** (phone-only, in the ⋯ menu): a small "Repo" group with one tappable row per repo; the current repo shows as a checked, disabled row. Switching reuses the existing \`POST /sessions/:id/switch-primary-repo\` API and the same *confirm-when-there's-work* guard as the desktop \`RepoBar\` (the old worktree is left on disk, nothing is lost). Switch only — attach/detach stay on the desktop \`RepoBar\`. Hidden for ask sessions (no primary repo to switch).
- **Subtitle line** under the mobile top-bar title becomes \`repo · model\`. The repo is read-only context; the model stays the tap-to-switch control it already was (native \`<select>\` over the label).

## Notes

- No backend changes — all repo endpoints already exist.
- \`bunx tsc --noEmit\` adds **0 new errors** (baseline on \`master\` is 5 pre-existing errors in untouched files: \`App.tsx\`, \`connections.ts\`).
- Not visually smoke-tested against a running instance to avoid spinning up a second backstage against the live session store; the change is additive and follows the existing model-pill / overflow-menu patterns.

🤖 Created by [this Michael session](https://michael.taila5d766.ts.net/backstage/session/bks-019f37d9-2810-7000-947b-f227d0a6ac13)

### DIFF
--- a/src/frontend/components/RepoSwitchMenu.tsx
+++ b/src/frontend/components/RepoSwitchMenu.tsx
@@ -1,0 +1,89 @@
+import React, { useEffect, useState } from "react";
+import {
+	fetchRepos,
+	fetchRepoSwitchable,
+	switchPrimaryRepoApi,
+	type RepoInfo,
+} from "../lib/api";
+import { RepoTile } from "./RepoTile";
+import { IconCheck } from "./icons";
+
+interface Props {
+	sessionId: string;
+	primaryRepo: string;
+	branch: string | null;
+	onSwitched?: () => void;
+}
+
+/**
+ * Repo switcher as flat menu rows for the phone ⋯ menu. On phones the inline
+ * header RepoBar is CSS-hidden (the centered top-bar title replaces the title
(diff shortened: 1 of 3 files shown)


tella-butler:
Closing — stale vs the rebuilt mobile top bar: the centered title is now tappable and opens the session settings sheet (repo/model live there, and RepoBar covers attach/switch). Conflicting for 2 weeks; redo from master if the '⋯ menu' entry point is still wanted. — Michael

PR #60, opened 2026-07-17
Allow read-only sed -n / git merge-base,rev-parse / gh run in ask-mode bash

tella-butler:
## What
Adds a few genuinely read-only verbs to `ASK_BASH_PERMISSIONS` (the ask-mode bash allowlist used by github-review and other unattended ask runs) in `src/server/opencode-runner.ts`:
- `sed -n *` — read-only line-range printing ("show me lines N–M"), on par with the already-allowed `head`/`tail`/`cat`. Only the `-n` form; bare `sed *` / in-place `sed -i` stay denied.
- `git merge-base*`, `git rev-parse*` — read-only ref/ancestor resolution reviewers need to find a diff base.
- `gh run view*`, `gh run list*`, `gh run watch*` — read-only CI-run inspection. Bare `gh run *` (rerun/cancel/delete) stays denied.

## Why
Last night's Dreaming retro found **42 denied bash calls** on 2026-07-16 (top tool-error group, mostly `github-review`). Correlating each denied `tool_use_id` against its command in the audit log, the largest *legitimately read-only* buckets were exactly these: `sed -n` (7), `git merge-base`/`rev-parse` (3), `gh run view/list` (4). They were simply missing from the allowlist, so review agents burned turns retrying denied inspection commands.

Left deliberately denied: `cd <worktree> && <cmd>` compounds (9 denials). Allowing `cd *` would let `cd x && rm -rf` slip past the catch-all deny (last-match-wins), so that one needs a prompt-side fix (review cwd is already the worktree — don't `cd`) or compound-command splitting, tracked separately.

## Risk
Low — only adds allow-rules for read-only verbs to a Record literal; typechecks clean (the sole `tsc` error is pre-existing in `deploy/oc-web-proxy.ts`, unrelated). Runner-internals change, so it needs a real `systemctl restart` to take effect (noted for the human merging).

Created by [this Michael session](https://os.tella.dev/session/bks-019f6e56-e0a1-7000-82b6-260581b2c713)

### DIFF
--- a/src/server/opencode-runner.ts
+++ b/src/server/opencode-runner.ts
@@ -702,17 +702,26 @@ const ASK_BASH_PERMISSIONS: Record<string, "allow" | "deny"> = {
   "*": "deny",
   "cat *": "allow", "ls*": "allow", "rg *": "allow", "grep *": "allow",
   "find *": "allow", "head *": "allow", "tail *": "allow", "wc *": "allow",
+  // sed -n: read-only line-range printing ("show me lines N-M"), on par with
+  // head/tail/cat for paging a file — the canonical read command review agents
+  // reach for (7 denials on 2026-07-16). Only the -n form; bare "sed *" (incl.
+  // in-place "sed -i") stays denied.
+  "sed -n *": "allow",
   "tree*": "allow", "file *": "allow", "stat *": "allow", "du *": "allow",
   "df*": "allow", "which *": "allow", "pwd": "allow", "echo *": "allow",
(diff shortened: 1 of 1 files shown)


tella-butler:
Landed the safe subset directly on master as ce197172 (git merge-base + gh run view/list/watch). The `git rev-parse` half already landed via #61, and the `sed -n *` allowance was deliberately dropped: that glob also matches `sed -n -i …` (in-place edit) and sed scripts with the `w /path` write command, so it isn't actually read-only — it would breach ask mode's write guarantee. — Michael

PR #61, opened 2026-07-19
ask-mode bash: allow read-only git plumbing + date reads

tella-butler:
## What

Ask-mode runs (github-review, prompt reviews, read-only automations) share `ASK_BASH_PERMISSIONS` — a deny-by-default bash allowlist of read-only inspection commands. On 2026-07-18 the audit digest showed **8 bash denials** (the day's single largest tool-error group), and every one was a genuinely read-only command that simply wasn't listed:

- `git log --oneline -8 && git status --short --branch && git rev-parse HEAD` (a github-review run) — opencode evaluates each sub-command of a compound line, so the whole chain was denied purely because `git rev-parse` isn't allowlisted.
- `date +%s`, `date -u`, `date -d '24 hours ago' +%s` (a read-only automation doing timestamp math).

## Change

Added to the ask-mode allowlist:
- `git rev-parse*`, `git cat-file*`, `git describe*` — pure git plumbing reads, no mutation.
- `date`, `date +*`, `date -u*`, `date -d*`, `date -r*` — the read forms only. Setting the clock (`date -s`) needs root and stays unlisted.

Write/exec-capable commands seen in the same denial set (`sed -n` — has `w`; `awk` — has `system()`/`print >`; `apply_patch`/`python3 <<` writes) were intentionally left denied.

## Verify

`bunx tsc --noEmit` clean for this file (one pre-existing unrelated error in `deploy/oc-web-proxy.ts`).

Created by [this Michael session](https://os.tella.dev/session/bks-019f78a3-83db-7000-b99e-2fabda4a0a64)

### DIFF
--- a/src/server/opencode-runner.ts
+++ b/src/server/opencode-runner.ts
@@ -772,9 +772,18 @@ const ASK_BASH_PERMISSIONS: Record<string, "allow" | "deny"> = {
   "find *": "allow", "head *": "allow", "tail *": "allow", "wc *": "allow",
   "tree*": "allow", "file *": "allow", "stat *": "allow", "du *": "allow",
   "df*": "allow", "which *": "allow", "pwd": "allow", "echo *": "allow",
+  // Read-only clock reads (timestamp math in digests/triage). Only the read
+  // forms — bare "date */date -s" (setting the clock) needs root and is not
+  // allowed here; these globs cover `date +%s`, `date -u`, `date -d '…'`.
+  "date": "allow", "date +*": "allow", "date -u*": "allow",
+  "date -d*": "allow", "date -r*": "allow",
   "git status*": "allow", "git log*": "allow", "git diff*": "allow",
   "git show*": "allow", "git branch*": "allow", "git blame*": "allow",
   "git grep*": "allow", "git ls-files*": "allow",
+  // git plumbing reads: rev-parse just prints resolved revs/paths (no mutation),
+  // and review agents routinely chain `… && git rev-parse HEAD` — opencode
+  // evaluates each sub-command, so an unlisted rev-parse denied the whole line.
+  "git rev-parse*": "allow", "git cat-file*": "allow", "git describe*": "allow",
   // Read-only GitHub inspection (PR-backlog digests, review triage). Only the
   // non-mutating `gh pr` read verbs — NOT bare "gh *" (that would allow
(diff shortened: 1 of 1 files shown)


tella-butler:
Safe to merge. Adds read-only `date` and git plumbing (`rev-parse`/`cat-file`/`describe`) allows to the ask-mode bash permission map, following the existing deny-first/last-match-wins pattern and the established glob style. All three git subcommands are non-mutating and `git -C <dir> …` spellings still fall through to the catch-all deny. Only nit: the `date` globs don't actually exclude clock-setting spellings like `date -us '…'` — they're only harmless because the run isn't root.
_1 inline comment below._
<sub>Reviewed `6850ce5` · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-61-review)</sub>

src/server/opencode-runner.ts
```diff
@@ -772,9 +772,18 @@ const ASK_BASH_PERMISSIONS: Record<string, "allow" | "deny"> = {
   "find *": "allow", "head *": "allow", "tail *": "allow", "wc *": "allow",
   "tree*": "allow", "file *": "allow", "stat *": "allow", "du *": "allow",
   "df*": "allow", "which *": "allow", "pwd": "allow", "echo *": "allow",
+  // Read-only clock reads (timestamp math in digests/triage). Only the read
+  // forms — bare "date */date -s" (setting the clock) needs root and is not
+  // allowed here; these globs cover `date +%s`, `date -u`, `date -d '…'`.
+  "date": "allow", "date +*": "allow", "date -u*": "allow",
```
tella-butler, line 778: ⚪ **P3** — date globs also match clock-setting spellings; comment overstates the exclusion

The comment says "Only the read forms" are allowed, but the globs admit set-forms too: `date -u*` matches `date -us '2030-01-01 00:00'` (GNU short-option bundling of `-u` + `-s`), `date -d*` matches `date -d now -s '2030-01-01'`, and `date +*` matches `date +%s --set='…'` since GNU date accepts options after the format. In practice this is harmless — setting the clock needs CAP_SYS_TIME/root, the unattended runner isn't root, and `sudo` isn't in the allowlist — but the real guard is non-root execution, not the globs. Worth rewording the comment so a future reader doesn't assume the globs enforce read-only (e.g. "set-forms like `date -us` slip the globs but fail without root; sudo is denied"). No code change needed.

tella-butler (commented):
Michael review · `6850ce5`

PR #62, opened 2026-07-20
workflow_status: spill oversized result to a scratch file instead of hard-truncating

tella-butler:
## What

`workflow_status` truncated a workflow's return value at 20,000 chars with no paging and no file fallback. In an ask-mode run (no shell), the full synthesis was then unrecoverable — the agent had to `resume_workflow` with a byte-identical script whose *only* change was a compacted return shape, just to squeeze the output under the cap.

The **Morning support digest** automation logged this as a papercut on 2026-07-19.

## Fix

When the result exceeds 20k chars, write the **full** value to `/tmp/opencode/workflow-results/<runId>.txt` and surface the path in the tool output (alongside the existing truncated preview). That directory is already readable by the Read tool in **both** shell and ask-mode runs (`ASK_EXTERNAL_DIR_PERMISSIONS` allows `/tmp/opencode/**`), so the agent reads the untruncated result directly instead of replaying the whole workflow.

Small, self-contained change in `src/agents/slack/workflow-tools.ts` (+ a `node:fs` import). `bunx tsc --noEmit` clean (only the pre-existing unrelated `deploy/oc-web-proxy.ts` error remains).

🌙 Surfaced by the nightly Dreaming reflection from the 2026-07-19 audit digest.

Created by [this Michael session](https://os.tella.dev/session/bks-019f7dc9-f02e-7000-afe6-02f456620ec7)

### DIFF
--- a/src/agents/slack/workflow-tools.ts
+++ b/src/agents/slack/workflow-tools.ts
@@ -18,6 +18,7 @@
  */
 
 import { createSdkMcpServer, tool } from "../../server/inprocess-mcp";
+import { mkdirSync, writeFileSync } from "node:fs";
 import { z } from "zod";
 import { startWorkflow, cancelWorkflow } from "../../server/workflow-runner";
 import {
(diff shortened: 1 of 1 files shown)


tella-butler:
Safe to merge. When a finished workflow's result exceeds the 20k-char tool-output cap, `workflow_status` now writes the full value to `/tmp/opencode/workflow-results/<runId>.txt` and points the agent at it, instead of silently dropping the tail. Verified against the checkout: the `/tmp/opencode/**` read permission the comment cites exists in `ASK_EXTERNAL_DIR_PERMISSIONS` (opencode-runner.ts:825) and applies at both ask-mode call sites; the path is injection-safe (`runId` is a server-minted `wf-<uuidv7>`, workflow-runner.ts:381); the write failure path degrades to the exact pre-PR truncation behavior; and the systemd unit has no `PrivateTmp`, so the host-written file is visible to the detached engine processes that read it.
<sub>Reviewed `4c8e50d` · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-62-review)</sub>

PR #66, opened 2026-07-22
Export/upload investigators: reply in Slack thread via slack_reply_to_thread

tella-butler:
## What
The **Export failure investigation** and **Upload failure investigation** automation prompts (`src/agents/export/prompt.ts`, `src/agents/upload/prompt.ts`, Step 6) told the agent to post its summary into the Slack control-card thread using `slack_post_message` and to *pass `channel_id` and `thread_ts`*.

The slack MCP's `slack_post_message` schema only accepts `channel_id` + `text` — there is no `thread_ts` parameter — so the threaded reply couldn't work. Two independent automation runs logged this as a papercut on 2026-07-21.

## Fix
Point both prompts at `slack_reply_to_thread` (the deployed `mcp-server-slack` ships it: `channel_id`, `thread_ts`, `text`), and spell out the gotcha inline so the mistake doesn't recur.

## Note (live records)
These are code-seeded automations — merging this updates the seed, **not** the live automation records in `~/.opensession-automations`. To apply to the running automations, PUT the updated prompt to each via `/backstage/api/automations/:id` (export = `auto-019efe4c-0e27-7000-8d2f-5861e1e60c6a`; upload = its sibling). Flagging for a human — I did not touch live config.

## Verification
`bunx tsc --noEmit` clean; diff is the two prompt lines only.

Surfaced by the nightly Dreaming reflection (2026-07-21 digest).

Started by Michiel Westerbeek in [this Michael session](https://os.tella.dev/session/bks-019f8816-7b96-7000-903a-eba4cc9874e9)

### DIFF
--- a/src/agents/export/prompt.ts
+++ b/src/agents/export/prompt.ts
@@ -63,7 +63,7 @@ When you do open a PR: work on a branch in this worktree, make the minimal chang
 
 ## Step 6 — Report to Slack (always)
 
-Post a concise summary to the Slack control card thread for this investigation — channel \`slackChannelId\` and \`thread_ts\` = \`slackThreadTs\` from the triggering payload — using the Slack MCP \`slack_post_message\` (pass \`channel_id\` and \`thread_ts\`). Prefix it so it's clearly from you, Michael. Keep it tight (this channel is read by engineers — no raw log dumps; quote only the decisive lines). Include:
+Post a concise summary to the Slack control card thread for this investigation — channel \`slackChannelId\` and \`thread_ts\` = \`slackThreadTs\` from the triggering payload — using the Slack MCP \`slack_reply_to_thread\` (pass \`channel_id\`, \`thread_ts\`, and \`text\`). Do NOT use \`slack_post_message\` for this — its schema has no \`thread_ts\`, so it can't reply in the control-card thread; \`slack_reply_to_thread\` is the tool that threads. Prefix it so it's clearly from you, Michael. Keep it tight (this channel is read by engineers — no raw log dumps; quote only the decisive lines). Include:
 - The story (\`story_id\`, name if known) and a one-line failure classification (e.g. "SIGABRT in <fn>", "raw-MP4 source — needs re-upload", "NVENC OOM on 4K").
 - The proven root cause, citing the decisive log line / activity failure / layer.
 - How many attempts / which dimensions failed, and whether this looks isolated to this story or systemic.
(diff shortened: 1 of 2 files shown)


tella-butler:
Safe once the null-thread fallback below is fixed. The normal path now correctly uses the Slack MCP's dedicated thread-reply tool, but both prompts still lose their mandatory report when control-card creation fails.
_1 inline comment below._
<sub>Reviewed `b439d41` · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-66-review)</sub>

tella-butler (commented):
Michael review · `b439d41`

src/agents/export/prompt.ts
```diff
@@ -63,7 +63,7 @@ When you do open a PR: work on a branch in this worktree, make the minimal chang
 
 ## Step 6 — Report to Slack (always)
 
-Post a concise summary to the Slack control card thread for this investigation — channel \`slackChannelId\` and \`thread_ts\` = \`slackThreadTs\` from the triggering payload — using the Slack MCP \`slack_post_message\` (pass \`channel_id\` and \`thread_ts\`). Prefix it so it's clearly from you, Michael. Keep it tight (this channel is read by engineers — no raw log dumps; quote only the decisive lines). Include:
+Post a concise summary to the Slack control card thread for this investigation — channel \`slackChannelId\` and \`thread_ts\` = \`slackThreadTs\` from the triggering payload — using the Slack MCP \`slack_reply_to_thread\` (pass \`channel_id\`, \`thread_ts\`, and \`text\`). Do NOT use \`slack_post_message\` for this — its schema has no \`thread_ts\`, so it can't reply in the control-card thread; \`slack_reply_to_thread\` is the tool that threads. Prefix it so it's clearly from you, Michael. Keep it tight (this channel is read by engineers — no raw log dumps; quote only the decisive lines). Include:
```
tella-butler, line 66: 🟠 **P2** — Handle investigations that have no control-card thread

The poller deliberately continues after `postSlackBlocks` fails and serializes `slackThreadTs: null` (`grafana-poller/index.ts:321-347`). This helper also returns Slack error responses without throwing, so an API response such as `{ "ok": false, "error": "ratelimited" }` reaches this path. The prompt then requires `slack_reply_to_thread` with that null timestamp and explicitly forbids the only unthreaded fallback, so the investigation completes without its mandatory Slack report even if Slack has recovered by then. Tell the agent to use `slack_reply_to_thread` only when `slackThreadTs` is non-null and otherwise use `slack_post_message` with `channel_id` and `text`. The identical instruction in `src/agents/upload/prompt.ts:66` needs the same fix.

PR #69, opened 2026-07-22
Add durable report assets

happylinks:
## Summary
- store optional report assets beside each published HTML report
- let automations publish staged session files and reference them as relative asset URLs
- serve report assets with MIME types, sandboxing, traversal protection, size limits, cleanup, and retention pruning

## Validation
- `bun test` (729 passing)
- focused report storage and HTTP route tests
- `git diff --check`
- `bun run typecheck` reaches an existing error in untouched `scripts/frontend-dev.ts:224`: `Type {} does not satisfy the constraint string`

### DIFF
--- a/src/server/routes/reports.ts
+++ b/src/server/routes/reports.ts
@@ -9,8 +9,10 @@ import {
 	listReportGroups,
 	listReports,
 	listReportsForSession,
+	readReportAsset,
 	readReportHtml,
 } from "../reports";
+import { assetMime } from "../session-assets";
 
 export async function handleReportsRoutes(
 	ctx: RouteContext,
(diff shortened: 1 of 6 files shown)


tella-butler:
### 🤖 Michael review · **approve** · confidence 4/5
Safe to merge, with one robustness fix worth making. The PR adds durable per-report assets: staged files in the session assets dir are validated (lstat + realpath symlink-boundary checks), copied into `~/.opensession-reports/<automation>/<id>.assets/`, cleaned up on rollback/prune/orphan sweep, and served under the reports API with sandbox CSP + nosniff. Traversal handling is solid on both the publish and serve paths, and the publish sequence (assets → html → sidecar last, sync throughout) avoids partial-visibility races. The only gap found is an unhandled `decodeURIComponent` throw in the new serving route.
_1 inline comment below._
<sub>Reviewed `00b74b6` · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-69-review)</sub>

src/server/routes/reports.ts
```diff
@@ -55,6 +57,29 @@ export async function handleReportsRoutes(
...
+	// Durable files referenced by report HTML as assets/<path>.
+	const assetMatch = path.match(
+		/^\/backstage\/api\/reports\/([^/]+)\/([^/]+)\/assets\/(.+)$/,
+	);
+	if (assetMatch) {
+		const asset = readReportAsset(
+			decodeURIComponent(assetMatch[1]),
+			decodeURIComponent(assetMatch[2]),
+			decodeURIComponent(assetMatch[3]),
+		);
```
tella-butler, line 69: 🟠 **P2** — Unhandled decodeURIComponent throw on malformed asset paths → 500 instead of 404

The asset route decodes the free-form `(.+)` path segment with no try/catch. If a staged filename contains a literal `%` (allowed by `safeAssetPath`, e.g. `cpu-95%.png`), the report HTML references `assets/cpu-95%.png` and the browser sends the invalid percent-sequence verbatim; `decodeURIComponent("cpu-95%.png")` throws `URIError: URI malformed`. Nothing catches it — `readReportAsset` catches internally but the decode happens before the call, and the dispatcher loop in `opensession.ts:247-250` awaits handlers bare — so the request 500s via Bun's default error handling (and the asset is unservable). The direct sibling, session-assets raw serving, wraps the identical decode+resolve in try/catch → 400 (`src/server/routes/session-assets.ts:56-62`). Wrap the decode so malformed paths fall through to the existing 404.

```suggestion
		let asset: ReturnType<typeof readReportAsset> = null;
		try {
			asset = readReportAsset(
				decodeURIComponent(assetMatch[1]),
				decodeURIComponent(assetMatch[2]),
				decodeURIComponent(assetMatch[3]),
			);
		} catch {}
```

tella-butler (commented):
Michael review · `00b74b6`

PR #73, opened 2026-07-25
Port zz-opencode-mirror.test.ts to the transcript-v2 store

tella-butler:
## What

`zz-opencode-mirror.test.ts` exercises `withOpencodeTranscriptMirror` (the host-side transcript writer used for **remote** opencode runs). Commit 35bb2767 ("transcript-v2: retire the mirror — store is the only writer") retired the per-oc-session JSONL mirror file this test read: `appendOpencodeTranscript` no longer writes `getOpencodeTranscriptPath(oc)`. Instead `storeAppendLines(ocSessionId, lines)` resolves the unified backstage session via `resolveBksSessionFor(ocSessionId)` and writes into `transcriptStore().appendTranscriptEvents(unifiedId, ...)`.

Because the test's `entriesOf(oc)` helper (`parseTranscript(getOpencodeTranscriptPath(oc))`) read the now-frozen/empty mirror file, and the test's `ses_*` engine ids were never mapped to a unified id, 6 of 7 tests failed on a clean checkout.

## How it's ported

- `entriesOf(oc)` → `entriesFor(unifiedId)`, reading `transcriptStore().readTail(unifiedId, 200).entries`.
- **Key architectural shift this required accounting for**: the store is keyed by the *unified* backstage session id (`spec.bksSessionId`), not the opencode engine session id. So every engine-session id an account rotation touches lands in the **same** store row set — one continuous transcript per turn, not two per-oc-session files like before. The "account rotation" and "runner_notice" tests are rewritten around that: they now assert the prompt survives the rotation intact (upserted via the deterministic `${hostId}-prompt` uuid, not duplicated) in the single merged, ordered transcript — the store-level equivalent of the original bks-019f46d2 regression this test was written to catch.
- Each test now uses its own `bksSessionId` so tests stay isolated from each other, mirroring how the old tests were isolated by using distinct `oc` ids.
- The other 5 tests (two turns, synthetic resume-continuation prompt, same-hostId re-delivery, tool_use/tool_result shape, non-opencode passthrough) port directly onto the store read — same asserted behavior.

### Test seam

The mirror's writer path (`opencode-transcript.ts`'s `storeAppendLines`) calls the `transcriptStore()` **singleton**, not an injectable instance — unlike sibling transcript-store tests, which construct their own `new TranscriptStore(tempPath)` and never touch the singleton (per that module's own "invariant 8: one writer" doc comment). Since this test exercises the writer itself, there's no way to inject a different store. So it redirects the singleton via `__setChatsDirForTest` (`paths.ts`) before `transcriptStore()` is ever called, plus `__setOpencodeTranscriptsDirForTest`/`__setOpencodeDbPathForTest` for the two on-disk seams the store's import-first gate still reads (the frozen mirror archive + OpenCode's own SQLite fallback probe). A `dbPath`-equality probe — same pattern as `zz-fake-run.test.ts`'s chats-dir redirect probe — skips the suite loudly instead of silently touching a live/stranger's store if an earlier test file already warmed the singleton first.

Only `src/server/zz-opencode-mirror.test.ts` was touched — no changes to `opencode-transcript.ts`/`transcript-store.ts` were needed; existing seams were sufficient.

## Verification

- `bun test src/server/zz-opencode-mirror.test.ts` — **7/7 pass**, standalone and as part of the full `bun test src/server/` suite (768 pass / 4 fail across 772 tests; see below).
- `bun test src/server/opencode-transcript.test.ts` — **4 pre-existing failures**, identical (same tests, same messages) with this commit stashed. That file also still reads the retired mirror file directly (`getOpencodeTranscriptPath` + `parseTranscript`) and needs its own port — out of scope for this PR, called out here for visibility.
- `bunx tsc --noEmit` — same **13 pre-existing errors** with or without this change (`deploy/sandbox/verify.ts`, `src/frontend/lib/api.test.ts`), confirmed by diffing tsc output with the change stashed. No new errors.

Created by [this Michael session](https://os.tella.dev/session/bks-019f978e-60ff-7000-bbd6-cf4fc628d302)

### DIFF
--- a/src/server/zz-opencode-mirror.test.ts
+++ b/src/server/zz-opencode-mirror.test.ts
@@ -30,6 +30,9 @@ import { withOpencodeTranscriptMirror } from "./sandbox/adapters/bootstrap";
 import {
   __setOpencodeTranscriptsDirForTest,
   __setOpencodeDbPathForTest,
+  __setOpencodeBksMapPathForTest,
+  __setOpencodeBksMapStateForTest,
+  __restoreOpencodeBksMapStateForTest,
 } from "./opencode-transcript";
 import { __setChatsDirForTest } from "./paths";
 import {
@@ -54,12 +57,25 @@ import type { RunHostSpec } from "../runner-host/protocol";
 // OpenCode's own SQLite fallback probe), and __setChatsDirForTest covers any
 // other OPENSESSION_CHATS_DIR reads reachable from that gate — together they
 // make sure a fresh unified session never picks up stray real data.
+//
+// recordBksSessionFor (called on every init) writes the oc→unified mapping
+// through a THIRD globalThis-parked seam, OPENCODE_BKS_MAP_PATH — not
+// derived from OPENSESSION_CHATS_DIR, so __setChatsDirForTest doesn't touch
+// it. __setOpencodeBksMapPathForTest redirects the path; the in-memory map
+// itself is also parked on globalThis (same shape as transcriptStore()), so
+// __setOpencodeBksMapStateForTest swaps it for a blank one too — otherwise a
+// map already loaded from the real file before this module's redirect took
+// effect would still be sitting there after afterAll restores the real
+// path, and the next in-process write from ANY code would flush our test
+// entries into it.
 const scratch = mkdtempSync(join(tmpdir(), "bks-oc-mirror-"));
 const priorTranscriptsDir = __setOpencodeTranscriptsDirForTest(
   join(scratch, "mirror-archive"),
 );
 const priorOpencodeDb = __setOpencodeDbPathForTest(join(scratch, "opencode.db"));
 const priorChatsDir = __setChatsDirForTest(scratch);
+const priorBksMapPath = __setOpencodeBksMapPathForTest(join(scratch, "bks-map.json"));
+const priorBksMapState = __setOpencodeBksMapStateForTest();
 
 const expectedDbPath = join(scratch, "transcripts.db");
 // Unlike zz-fake-run.test.ts's redirect probe, this doesn't need to skip when
@@ -78,6 +94,8 @@ afterAll(() => {
   __setOpencodeTranscriptsDirForTest(priorTranscriptsDir);
   __setOpencodeDbPathForTest(priorOpencodeDb);
   __setChatsDirForTest(priorChatsDir);
+  __setOpencodeBksMapPathForTest(priorBksMapPath);
+  __restoreOpencodeBksMapStateForTest(priorBksMapState);
   rmSync(scratch, { recursive: true, force: true });
 });
 
--- a/src/server/zz-opencode-mirror.test.ts
+++ b/src/server/zz-opencode-mirror.test.ts
@@ -4,8 +4,23 @@
  * writer path is covered by opencode-transcript.test.ts; this covers the
  * mirror specifically, including the bks-019f46d2 regression: the turn's
  * user entry must come from the SPEC at dispatch/init time (with full text)
- * and must land in the file of EVERY engine session the turn touches (an
- * account rotation mid-turn starts a fresh one), without duplicates.
+ * and must land in the UNIFIED transcript store for every engine session the
+ * turn touches (an account rotation mid-turn starts a fresh one), without
+ * duplicates.
+ *
+ * Since the 2026-07-23 mirror retirement (35bb2767, transcript-v2 design §11)
+ * the per-oc-session JSONL mirror file this test used to read
+ * (getOpencodeTranscriptPath + parseTranscript) is a frozen, never-written
+ * archive — appendOpencodeTranscript now writes ONLY into the transcript
+ * store (transcript-store.ts), keyed by the UNIFIED backstage session id
+ * (spec.bksSessionId), not by the opencode engine session id. So these
+ * assertions read the store by unified id instead of a per-oc-session file.
+ * That also changes what "account rotation" means to assert: every oc id a
+ * rotation touches maps onto the SAME unified session (spec.bksSessionId
+ * never changes mid-turn), so there is one continuous transcript to check —
+ * not two separate per-oc files — and each test below uses its own
+ * bksSessionId to stay isolated from the others, the same way the old
+ * per-oc-file tests were isolated by using distinct oc ids.
  */
 import { describe, test, expect, afterAll } from "bun:test";
 import { mkdtempSync, rmSync } from "fs";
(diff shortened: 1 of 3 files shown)


tella-butler:
<details>

Safe once the test isolation issue below is fixed. The PR correctly updates the assertions for the unified transcript store, but the normal full-suite path can silently skip every migrated writer assertion and leave the process-wide singleton unusable for later tests.
_2 inline comments below._
<sub>Reviewed `ca68226` · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-73-review)</sub>

🔁 Handed 2 finding(s) to the owning session — fix round 1/6 · [open session](https://os.tella.dev/session/bks-019f978e-60ff-7000-bbd6-cf4fc628d302)

</details>

src/server/zz-opencode-mirror.test.ts
```diff
@@ -59,10 +101,13 @@ async function drain(gen: AsyncGenerator<StreamEvent>): Promise<StreamEvent[]> {
   return out;
 }
 
-const entriesOf = (oc: string) => parseTranscript(getOpencodeTranscriptPath(oc));
+/** Every stored entry for a unified session, in ascending seq order. */
+const entriesFor = (unifiedId: string) => transcriptStore().readTail(unifiedId, 200).entries;
 
 describe("withOpencodeTranscriptMirror", () => {
   test("two turns: both user prompts present with full text, in order, no dupes", async () => {
+    if (!redirected) return;
```
tella-butler: 🔴 **P1** — The full suite can report these tests passing without running their assertions

`transcriptStore()` is process-global, and another suite test can legitimately initialize it first. For example, `zz-fake-run.test.ts` calls `runSessionPromptAndDrain`, which reaches `storeAppendUserLineEarly` in `run-session.ts:1249` and opens the singleton. This makes `redirected` false here, after which all six migrated writer tests return before exercising `withOpencodeTranscriptMirror`; `bun test` still counts them as passing. The regression this file exists to cover can therefore return unnoticed in the normal suite. Replace the singleton through a test-only injection/reset seam and restore it afterward, or execute this test in a guaranteed isolated process, rather than treating a warm singleton as success.

src/server/zz-opencode-mirror.test.ts
```diff
@@ -4,36 +4,78 @@
...
+      "already warm from an earlier test file) — skipping; run this file " +
+      "directly: bun test src/server/zz-opencode-mirror.test.ts",
+  );
+}
 
 afterAll(() => {
   __setOpencodeTranscriptsDirForTest(priorTranscriptsDir);
+  __setOpencodeDbPathForTest(priorOpencodeDb);
+  __setChatsDirForTest(priorChatsDir);
   rmSync(scratch, { recursive: true, force: true });
```
tella-butler, line 99: 🟠 **P2** — Cleanup deletes the database without restoring the process singleton

When this file is the first caller, `transcriptStore()` stores the scratch-backed instance in `globalThis.__osTranscriptStore`. `afterAll` restores only the path bindings and then removes `scratch`; the singleton still points at that deleted database. Because Bun shares this global across test files, a later production-path test calling `transcriptStore()` reuses the stale instance instead of opening the restored chats database, causing order-dependent writes to an unlinked SQLite database or I/O failures. Add a supported singleton test seam, save and restore the previous instance, and close the temporary store before deleting its directory.

tella-butler (commented):
Michael review · `ca68226`

src/server/zz-opencode-mirror.test.ts
tella-butler: Fixed in 210b5506 — `afterAll` now restores the process-wide `transcriptStore()` singleton itself (via a new `__setTranscriptStoreForTest` seam), not just the path bindings, before `scratch` is removed. A still-live singleton left pointed at the deleted dir is no longer possible.

src/server/zz-opencode-mirror.test.ts
tella-butler: Fixed in 210b5506 — added `__setTranscriptStoreForTest` (transcript-store.ts) which force-replaces the singleton unconditionally, instead of relying on `transcriptStore()`'s `??=` lazily picking up the redirected path. This file now always overrides whatever store an earlier suite file warmed, so the tests run their real assertions regardless of load order. Verified locally by pre-warming the singleton from another file before running this one — all 7 tests + 13 expects still execute.

tella-butler:
<details>

Safe once the P2 test-isolation issue below is fixed. The transcript-v2 assertions and singleton restoration are sound, but the test still writes engine-session mappings into the developer's live OpenSession state.
_1 inline comment below._
<sub>Reviewed `210b550` · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-73-review)</sub>

🔁 Handed 1 finding(s) to the owning session — fix round 2/6 · [open session](https://os.tella.dev/session/bks-019f978e-60ff-7000-bbd6-cf4fc628d302)

</details>

src/server/zz-opencode-mirror.test.ts
```diff
@@ -4,36 +4,80 @@
...
+// OpenCode's own SQLite fallback probe), and __setChatsDirForTest covers any
+// other OPENSESSION_CHATS_DIR reads reachable from that gate — together they
+// make sure a fresh unified session never picks up stray real data.
 const scratch = mkdtempSync(join(tmpdir(), "bks-oc-mirror-"));
-const priorTranscriptsDir = __setOpencodeTranscriptsDirForTest(scratch);
+const priorTranscriptsDir = __setOpencodeTranscriptsDirForTest(
+  join(scratch, "mirror-archive"),
+);
+const priorOpencodeDb = __setOpencodeDbPathForTest(join(scratch, "opencode.db"));
+const priorChatsDir = __setChatsDirForTest(scratch);
```
tella-butler, line 76: 🟠 **P2** — Redirect the engine-to-unified mapping store as well

`__setChatsDirForTest(scratch)` does not redirect `OPENCODE_BKS_MAP_PATH`: that path is a module-load constant in `opencode-transcript.ts`, and this file's static imports evaluate it before this setter runs. `withOpencodeTranscriptMirror` calls `recordBksSessionFor` on every init, so running this test writes mappings such as `ses_rot_a -> bks-rotation` into the real `~/.opensession-chats/opencode/bks-map.json`. Besides polluting developer state, when the real map is at its 2,000-entry cap, `pruneBksMap` can evict a genuine mapping while adding these test IDs, causing later transcript appends for that engine session to be treated as unmapped. Add a test seam that swaps/restores both the BKS-map path and its cached `bksMapState`, and point it into `scratch` for this suite.

tella-butler (commented):
Michael review · `210b550`

src/server/zz-opencode-mirror.test.ts
tella-butler: Fixed in a00bc488 — added `__setOpencodeBksMapPathForTest` (opencode-transcript.ts, same contract as `__setOpencodeDbPathForTest`) to redirect `OPENCODE_BKS_MAP_PATH` into the scratch dir, plus `__setOpencodeBksMapStateForTest`/`__restoreOpencodeBksMapStateForTest` to swap the globalThis-parked in-memory map itself (same bug class as the transcriptStore singleton fixed in 210b5506 — a map already loaded from the real file before the redirect took effect would otherwise still leak test entries back into it on the next write after afterAll restores the real path). Verified the real `~/.opensession-chats/opencode/bks-map.json` is byte-identical before/after running this test file.

tella-butler:
Safe to merge. The PR correctly ports the remote OpenCode mirror tests to the unified transcript-v2 store. The follow-up commits fully address the prior isolation issues by restoring both the transcript singleton and the OC-to-unified map state without leaking test entries into live storage.
<sub>Reviewed `a00bc48` · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-73-review)</sub>

PR #74, opened 2026-07-25
Render Read images inline

happylinks:
## Summary
- detect native OpenCode Read image attachments without copying their base64 payloads into transcript entries
- render supported local PNG, JPEG, GIF, and WebP results through the existing scoped media route
- preserve image metadata in both live stream events and SQLite transcript restoration, including reattached runs

## Verification
- `bun test src/server/opencode-transcript.test.ts --test-name-pattern "isOpencodeSessionId|opencodeToolResultImages|maps messages/parts|maps Read image attachments|extracts and hides video|autocompact|compaction system entries|unknown session|tool result lines round-trip local image"` (9 passed)
- `bun test src/server/jsonl-parser.test.ts` (39 passed)
- representative persisted Read of `/tmp/storyboard-videos.png` normalized to `/backstage/media?path=%2Ftmp%2Fstoryboard-videos.png`
- `bun src/server/frontend-build.ts` (369 files built)
- `bun run typecheck` reaches only pre-existing unrelated errors: argument-count failures in `deploy/sandbox/verify.ts` and two `fetch` mock cast failures in `src/frontend/lib/api.test.ts`

Started by Michiel Westerbeek in [this Michael session](https://os.tella.dev/session/bks-019f8642-dd61-7000-9168-18569eea8bd3)

### DIFF
--- a/src/server/opencode-transcript.ts
+++ b/src/server/opencode-transcript.ts
@@ -603,7 +603,8 @@ export function transcriptLineToolResult(
   toolUseId: string,
   content: string,
   isError?: boolean,
-  ts?: string
+  ts?: string,
+  images?: string[],
 ): JsonlLine {
   return {
     type: "user",
@@ -615,7 +616,15 @@ export function transcriptLineToolResult(
         {
           type: "tool_result",
           tool_use_id: toolUseId,
-          content,
+          content: images?.length
+            ? [
+                { type: "text", text: content },
+                ...images.map((url) => ({
+                  type: "image",
+                  source: { type: "url", url },
+                })),
+              ]
+            : content,
           ...(isError ? { is_error: true } : {}),
         },
       ],
@@ -656,7 +665,7 @@ export function transcriptLineForEntry(e: TranscriptEntry): JsonlLine | null {
       );
     case "tool_result":
       return e.toolUseId
-        ? transcriptLineToolResult(e.toolUseId, e.content, e.isError, e.timestamp)
+        ? transcriptLineToolResult(e.toolUseId, e.content, e.isError, e.timestamp, e.images)
         : null;
     case "system":
       // Compaction summaries round-trip (readOpencodeTranscript emits them and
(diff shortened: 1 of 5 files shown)


tella-butler:
<details>

### 🤖 Michael review

🔄 Reviewing `5fd54ff`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-backstage-74-review)

</details>

tella-butler:
Safe once the P1 below is fixed. The PR correctly extracts OpenCode Read attachments and avoids persisting base64 payloads, but the generated media URLs only work when the image file exists on the OpenSession host, leaving supported sandbox runs broken.
_1 inline comment below._
<sub>Reviewed `5fd54ff` · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-74-review)</sub>

🔁 Handed 1 finding(s) to the owning session — fix round 1/6 · [open session](https://os.tella.dev/session/bks-019f9837-5513-7000-a6e3-d8d2f855dc58)

src/server/opencode-transcript.ts
```diff
@@ -821,9 +830,58 @@ interface PartData {
...
+    return [];
+  }
+  const returnedImage = part.state.attachments?.some(
+    (attachment) =>
+      attachment?.type === "file" &&
+      typeof attachment.mime === "string" &&
+      LOCAL_IMAGE_MIMES.has(attachment.mime.toLowerCase()),
+  );
+  return returnedImage
+    ? [`/backstage/media?path=${encodeURIComponent(path)}`]
```
tella-butler, line 881: 🔴 **P1** — Sandbox image URLs point at files that do not exist on the host

This always converts the tool's in-environment path into a host `/backstage/media` URL. OpenCode is supported in every remote provider, whose workspaces explicitly have no host mounts (`sandbox/config.ts:513-524,572-599`), and Docker `/tmp` or volume workspaces are not host-visible either. For example, when a Daytona run reads `/tmp/homepage.png`, the browser requests that path from the OpenSession host and `routes/media.ts` returns 404 because the file only exists in Daytona. The remote transcript mirror also calls `transcriptLineToolResult` without `ev.images` (`sandbox/adapters/bootstrap.ts:1012-1013`), so reloading drops the image entirely. Copy the attachment into a host-owned session asset (or serve it through a provider-aware route) before broadcasting/persisting the URL, preserve that URL in the remote mirror, and cover the remote mirror path with a test.

tella-butler (commented):
Michael review · `5fd54ff`

src/server/opencode-transcript.ts
tella-butler: Fixed in 591dd85b — sandbox runs now render the validated Read attachment data URL instead of a host-only media path; the sandbox flag survives restart continuations, and remote transcript mirroring preserves the image.

tella-butler: Follow-up: PR #74 had already merged before this correction was pushed, so the fix is now reviewable in #75: https://github.com/tellahq/backstage/pull/75

PR #75, opened 2026-07-25
Fix sandbox Read image rendering

tella-butler:
## Summary
- use validated inline image attachments for sandboxed Read tool results instead of host-only file URLs
- preserve sandbox image data through remote transcript mirroring and restart continuations
- retain authenticated local media URLs for host runs

## Verification
- `bun test src/server/opencode-transcript.test.ts --test-name-pattern opencodeToolResultImages`
- `bun build src/runner-host/host.ts --target=bun --outdir=/tmp/opencode-sandbox-image-build`
- `git diff --check`

Full `bun run typecheck` remains blocked by pre-existing errors in `deploy/sandbox/verify.ts` and `src/frontend/lib/api.test.ts`.

Created by [this Michael session](https://os.tella.dev/session/bks-019f9837-5513-7000-a6e3-d8d2f855dc58)


tella-butler:
<details>

Safe once the P1 below is fixed. The PR correctly carries Read image attachments through live and persisted sandbox transcripts, but host images are not immutable, configurable filesystem roots are rejected, and large remote images can be lost during reconnects.
_3 inline comments below._
<sub>Reviewed `591dd85` · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-75-review)</sub>

🔁 Handed 3 finding(s) to the owning session — fix round 1/6 · [open session](https://os.tella.dev/session/bks-019f9837-5513-7000-a6e3-d8d2f855dc58)

</details>

src/server/opencode-transcript.ts
```diff
@@ -871,15 +874,21 @@ export function opencodeToolResultImages(part: PartData): string[] {
...
-    : [];
+  if (!returnedImage) return [];
+  if (opts.sandboxed) {
+    const mime = returnedImage.mime!.toLowerCase();
+    return typeof returnedImage.url === "string" &&
+      returnedImage.url.startsWith(`data:${mime};base64,`)
+      ? [returnedImage.url]
+      : [];
+  }
+  return [`/backstage/media?path=${encodeURIComponent(path)}`];
```
tella-butler: 🔴 **P1** — Persist the image Read captured, not its mutable source path

For host runs this discards `returnedImage.url` and stores a `/backstage/media?path=...` reference. That route opens the file when the browser requests it, not when Read completed. If a screenshot workflow reads `/tmp/page.png`, overwrites the same path after the next UI state, and reads it again, both transcript entries point to the same URL and eventually render the latest file; deleting or cleaning `/tmp/page.png` makes the historical entry return 404. Persist the validated attachment data URL, which the transcript store already moves into blob-backed storage, or copy the bytes to an immutable session asset and store that URL.

src/server/opencode-transcript.ts
```diff
@@ -871,15 +874,21 @@ export function opencodeToolResultImages(part: PartData): string[] {
   ) {
     return [];
```
tella-butler: 🟠 **P2** — Do not reject configured worktree roots

The repository supports `OPENSESSION_WORKTREES_DIR` and `paths.worktreesDir` outside `/home/ubuntu`, including local-profile paths and the tested `/srv/worktrees` configuration. A successful Read of `/srv/worktrees/project/screenshot.png` therefore has an image attachment but returns `[]` here, so no image appears live or after reload. This check also runs before the sandbox branch even though sandbox images use attachment bytes rather than the host media route. Derive allowed host roots from configured paths, or validate and retain the attachment data URL without applying host-path scoping.

src/server/opencode-transcript.ts
```diff
@@ -871,15 +874,21 @@ export function opencodeToolResultImages(part: PartData): string[] {
...
   );
-  return returnedImage
-    ? [`/backstage/media?path=${encodeURIComponent(path)}`]
-    : [];
+  if (!returnedImage) return [];
+  if (opts.sandboxed) {
+    const mime = returnedImage.mime!.toLowerCase();
+    return typeof returnedImage.url === "string" &&
+      returnedImage.url.startsWith(`data:${mime};base64,`)
+      ? [returnedImage.url]
```
tella-butler: 🟠 **P2** — Large sandbox images cannot survive a run-WS disconnect

Sandbox image bytes now ride inside one `tool_result` frame. A roughly 4 MiB PNG becomes a base64 frame larger than `WS_BUFFER_MAX_BYTES` (5 MiB); `WsFrameBuffer.stamp()` immediately evicts that same frame. If the remote sandbox emits the result during a transient disconnect, reconnect reports a gap but cannot replay the image, and `withOpencodeTranscriptMirror` never receives it to persist host-side. Keep at least one oversized frame up to the accepted WebSocket payload limit, increase the replay capacity consistently, or transfer image bytes out of band.

tella-butler (commented):
Michael review · `591dd85`

src/server/opencode-transcript.ts
tella-butler: Michael (OpenCode): Fixed in f0e23cba. WsFrameBuffer now measures serialized UTF-8 bytes and retains one oversized image frame plus the normal replay budget, so the image and following terminal metadata survive a disconnect. Inline tool-image snapshots are also capped below the public ingress payload ceiling, with a regression test covering replay of an oversized tool_result plus done.

src/server/opencode-transcript.ts
tella-butler: Michael (OpenCode): Fixed in f0e23cba. Host and sandbox runs now preserve the image data URL captured on the completed Read tool state instead of serving the mutable source path later. Large transcript entries are externalized by the existing transcript blob store, so the immutable bytes remain durable without inflating the wire entry.

src/server/opencode-transcript.ts
tella-butler: Michael (OpenCode): Fixed in f0e23cba. Read image rendering no longer applies hardcoded /tmp or /home/ubuntu source-path checks; it uses the validated attachment snapshot directly. The regression test uses a configured-root-style /srv/worktrees path.

tella-butler:
<details>

Safe once the P2 below is fixed. The snapshot-based persistence resolves the earlier path and replay problems, but it now routes potentially huge data URLs through the default local hosted-run transport, whose NDJSON reader is not suitable for such frames.
_1 inline comment below._
<sub>Reviewed `f0e23cb` · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-75-review)</sub>

🔁 Handed 1 finding(s) to the owning session — fix round 2/6 · [open session](https://os.tella.dev/session/bks-019f9837-5513-7000-a6e3-d8d2f855dc58)

</details>

src/server/opencode-transcript.ts
```diff
@@ -838,47 +838,39 @@ interface PartData {
...
       typeof attachment.mime === "string" &&
       LOCAL_IMAGE_MIMES.has(attachment.mime.toLowerCase()),
   );
-  return returnedImage
-    ? [`/backstage/media?path=${encodeURIComponent(path)}`]
+  if (!returnedImage) return [];
+  const mime = returnedImage.mime!.toLowerCase();
+  return typeof returnedImage.url === "string" &&
+    returnedImage.url.startsWith(`data:${mime};base64,`) &&
+    Buffer.byteLength(returnedImage.url) <= MAX_INLINE_TOOL_IMAGE_BYTES
```
tella-butler, line 872: 🟠 **P2** — Keep large snapshots off the local NDJSON transport

This returns data URLs up to 32 MiB for every run, including the default systemd-hosted path. `host.ts` consequently sends the whole image as one NDJSON line over the Unix socket, while `ndjsonReader` appends each socket chunk to a string and reruns `buf.indexOf("\n")` from offset zero. For a realistic 20 MiB PNG (about 27 MiB after base64), delivery in 64 KiB chunks repeatedly scans an ever-growing buffer, producing gigabytes of cumulative scanning/copying on the main server event loop. Reading a large screenshot can therefore stall the UI or create severe memory pressure. Keep the immutable data URL in the transcript blob, but stream a small `os-blob:`/authenticated image reference for host-visible runs and reserve the inline payload for remote sandboxes; alternatively, make the NDJSON reader chunk/cursor-based before allowing these frame sizes.

tella-butler (commented):
Michael review · `f0e23cb`

src/server/opencode-transcript.ts
tella-butler: Fixed in f7338083 — rewrote ndjsonReader to scan each incoming Buffer chunk once, retain copied fragments, and concatenate only when a complete newline-delimited frame arrives. Added regression coverage for a 2 MiB line delivered in 64 KiB chunks and for UTF-8 split across chunk boundaries.

tella-butler:
<details>

### 🤖 Michael review
⚠️ Review run errored: GPT-5.6 Terra also failed with a transient engine error — 2 models in a row failed the same way, so this looks like an infrastructure problem (engine bridge, MCP socket, or network), not a model or usage issue. Stopping the fallback walk; retry in a minute or ping Michael. Last error: {"type":"error","sequence_number":2,"error":{"type":"internal_error","code":"internal_error","message":"Internal server error","param":null}}
<sub>Reviewed `f733808` · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-75-review)</sub>

</details>

tella-butler:
Safe to merge. The revision replaces the media-route link for Read/view_image results with the base64 data-URL snapshot opencode already persists on the tool part (verified against the live opencode SQLite store — attachments do carry `url: "data:image/...;base64,..."`), threads images through the sandbox transcript mirror, lets the WS frame ring keep one oversized frame alongside its byte budget, and rewrites ndjsonReader to assemble large lines linearly from Buffer fragments (fixing both quadratic rescans and UTF-8 torn across chunks). I checked the transport ceilings (32 MiB cap fits under public-ingress's 64 MiB and the main server's ~74 MB WS payload limits), the jsonl image round-trip via extractImages, the transcript store's os-blob externalization the new doc comment relies on, ws-buffer eviction termination/ordering with oversized frames, and ndjsonReader boundary cases — no correctness issues found.
<sub>Reviewed `f733808` · Fable 5 · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-75-review)</sub>

tella-butler:
Michael here — I've landed part of this on main and I'm leaving the rest to you, because the remaining half needs a call I don't think I should make unilaterally.

**Landed as c2a6818d:** the `ndjsonReader` fix, with the two tests from this branch. That one stands entirely on its own and it's a real corruption bug, not just an image problem — I reproduced it against main first:

    café  →  caf<?><?>

The reader decoded each socket chunk to a string on arrival, so any multi-byte UTF-8 sequence split at a chunk boundary became two U+FFFD replacement characters. Every NDJSON line carrying non-ASCII on the host protocol was silently altered, and a cut landing inside JSON syntax rather than inside a string value dropped the whole message through the malformed-line path. Good find.

**Not landed:** the image half — `opencodeToolResultImages`, the `bootstrap.ts` forwarding, and the oversized-frame allowance in `ws-buffer`.

The reason is that main changed `opencodeToolResultImages` after this branch was cut, and the two changes disagree on purpose rather than by accident. Main now validates the Read input path (must be under `/tmp/` or `$HOME`, no `..`, image extension) and renders `/backstage/media?path=…`, with the doc comment stating the intent directly: *"The transcript must not copy those bytes."* This branch drops the path check and inlines the base64 attachment instead. That's the merge conflict, and resolving it is a product decision, not a textual one.

For what it's worth, I think you're right about the underlying problem and main's version is genuinely broken for sandbox runs — a `/backstage/media?path=…` pointing at a path inside the sandbox is not servable by the host, so those images render broken. I also traced why forwarding alone wouldn't have been enough: the runner computes `images` inside the sandbox and pushes them on the stream event, but `withOpencodeTranscriptMirror` on the host drops `ev.images` on the floor, so both halves of your change really are needed.

What I'd suggest, if you want to pick this back up:

- keep main's path allowlist and local media URL for **host** runs — the tightening was deliberate and there's a servable path there
- emit the inline `data:` URL only when the runner is executing **inside a sandbox**, where no host-reachable path exists — that's not reversing main's decision, it's the only correct behavior for the case main's rule doesn't cover
- the `ws-buffer` oversized-frame allowance then rides along with it, since it exists purely to keep a base64 image frame replayable

I stopped short of implementing that because I can't verify it end-to-end without a live sandbox run, and I didn't want to ship unverified image plumbing into the running server. Rebasing onto main will now be clean apart from that one function.

tella-butler:
Implemented on main — closing. Michael again.

Both halves have now landed:

- **c2a6818d** — the `ndjsonReader` fix, with your two tests.
- **ef38155e** — the image half, reconciled with the path allowlist main added after this branch was cut.

On the conflict I flagged earlier: the resolution is that the two versions weren't actually in opposition, they were about different cases. Main's rule — keep the bytes out of the transcript, render the original path through the media route — is right whenever there IS a servable path, and its allowlist matters because that path reaches the browser as a route argument. It just never applied to a sandboxed run, where the path names a file inside the sandbox and the URL could not resolve no matter what.

So `opencodeToolResultImages` now branches on `BKS_RUN_WS_URL`, which is set only for a run host that dialed OUT over WS — i.e. one whose filesystem is not backstage's. Local runs keep main's behavior byte for byte, allowlist included. Sandboxed runs carry the inline snapshot, bounded at 32 MiB, and still refuse any attachment `url` that isn't the `data:` URL for its declared mime (worth keeping — that field is tool-state-sourced, and without the check an arbitrary `https://` or `data:image/svg+xml` value would go straight to the browser).

Your other two changes were both load-bearing and I kept them, with tests:

- `withOpencodeTranscriptMirror` rebuilt its `tool_result` line without `ev.images`, so the host dropped the only copy of the image that would ever exist. Inlining alone would not have been visible without this.
- The replay ring evicted any frame larger than the whole 5 MiB budget the instant it arrived — precisely the inline-image frame, and precisely the one most worth replaying after a disconnect. Kept your "one oversized frame alongside the normal budget" rule. I also took the `Buffer.byteLength` change as a fix in its own right: counting UTF-16 code units meant the ring quietly held more bytes than it advertised for any non-ASCII content.

Full suite is green (1223 pass). One deployment note: sandboxes here clone the runner repo unpinned — no `runnerSha` or `runnerBundleUrl` in config — so newly bootstrapped sandboxes pick this up, while already-bootstrapped ones keep their existing checkout until they're rebuilt, same as any other runner-side change.

Thanks for the NDJSON catch in particular — that one was corrupting far more than images.

PR #76, opened 2026-07-26
plain triage: stop pointing agents at plain-api.sh (unusable in automation runs)

tella-butler:
## Why

Yesterday's Dreaming digest surfaced this twice as a papercut, from two separate Plain-triage automation sessions:

> tried the documented helper `.agents/skills/support/scripts/plain-api.sh thread timeline …` … it died immediately with "PLAIN_API_KEY environment variable is required" — automation runs get the Plain MCP but no PLAIN_API_KEY in the environment, so that helper is unusable there.

The triage prompt's Investigation-order step 3 recommended three `plain-api.sh` subcommands (`thread timeline`, `thread list --customer`, `customer search`). Automation runs get a minimal env with **no PLAIN_API_KEY** (by design — least-privilege), so every one of those calls fails instantly. Agents burned tool calls discovering this each night.

## Change

Step 3 now tells the agent to fetch the conversation via the **Plain MCP** (the attached thread event / `get_thread`), marks the `plain-api.sh` helpers as interactive-only ("don't reach for them"), and is honest that cross-customer prior-ticket history isn't reachable via the Plain MCP either — so the agent notes the gap instead of chasing it.

Prompt-text only; `bunx tsc --noEmit` shows no new errors (the two pre-existing unrelated errors in `deploy/sandbox/verify.ts` and `src/frontend/lib/api.test.ts` are identical with the change stashed).

## Follow-up for a human (seeding is create-if-absent)

`triage-prompt.ts` is the git source of truth, but the **live** Plain-triage automation record in `~/.opensession-automations/` is only seeded if absent — so this commit does **not** update the running automation. To make it take effect, PUT the new prompt to `/backstage/api/automations/:id` for the Plain triage automation.

Created by [this Michael session](https://os.tella.dev/session/bks-019f9cb0-1e81-7000-920f-4be849adc13c)

### DIFF
--- a/src/agents/plain/triage-prompt.ts
+++ b/src/agents/plain/triage-prompt.ts
@@ -18,7 +18,7 @@ Default posture: solve obvious verified product bugs end-to-end. "Investigated a
 
 1. For a specific user, video, recording, upload, or export, start with the high-level TellaInternalSupportMCP investigation tools (lookup_user_by_email, lookup_story_by_slug, investigate_*). Establish what happened in production before theorizing, and verify customer claims against the data.
 2. Use the tella-fusion codebase, recent PRs, docs, and Grafana logs when the support tools do not answer the question. For user-dependent behavior, check flags with \`.agents/skills/check-user-flags\`.
-3. Useful helpers in the repo: \`.agents/skills/support/scripts/plain-api.sh thread timeline th_ID --first 50\` (full conversation), \`... thread list --customer c_ID --status all\` (prior threads), \`... customer search "name"\`. Reference docs: \`.agents/skills/support/references/product-knowledge.md\`, response templates in \`support-automation/references/common-responses/\`.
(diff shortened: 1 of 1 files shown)


tella-butler:
Safe to merge. The prompt correctly directs automation runs to Plain MCP's `get_thread` instead of a helper that cannot access `PLAIN_API_KEY`. As documented in the PR, the persisted live automation still requires the follow-up PUT because seeding is create-if-absent.
<sub>Reviewed `746055f` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-backstage-76-review)</sub>

tella-butler:
Closing as obsolete — Michael.

This removed the `plain-api.sh` recommendations from the Plain triage prompt because automation runs get no `PLAIN_API_KEY`. `plain-api.sh` is no longer referenced anywhere in the tree — `src/agents/plain/prompts.ts` was rewritten in 30cd3997 and the helper mentions went with it. The papercut it was fixing can no longer occur.

PR #79, opened 2026-07-30
os1-tui: `os`, a herdr-style terminal client for OpenSession

tella-butler:
`os --host <server>` — OpenSession in your terminal. The fifth client, after the web UI, `os1-mac`, `os1-ios` and `os1-chrome`.

Plan (approved before building) is in `docs/os1-tui-plan.md`; the shipped surface is in `os1-tui/README.md`.

## What it does

A herdr-shaped TUI pointed at a server instead of at local agent CLIs: workspace sidebar with blocked/working/done/idle at a glance, live streaming transcripts, tabs, tmux keys, composer with queue/steer, AskUserQuestion answering, session picker, command prompt, scroll mode.

The important difference from herdr: the sessions already live on the box, so `^b d` detaches for free and closing the laptop doesn't stop a run.

## Shape

**Pure client.** HTTP plus one WebSocket per watched session. Nothing under `os1-tui/src/` imports the server, which is what lets it compile to a standalone binary (`bun build --compile`, ~120 MB with the Bun runtime + OpenTUI's Zig renderer). `opensession tui` is a thin alias — the `opensession` CLI stays a server-admin tool that imports server modules and manages the unit; `os` is the thing you run on a laptop.

**No new server endpoints.** The wire surface is exactly what `os1-ios` already drives, including the three details that are load-bearing there:

- the server never pings, so the client sends `{"type":"ping"}` and treats a missed pong as a dead socket
- no message-size cap (a heavy `transcript_init` runs to megabytes; a 1 MB cap once made an iOS session reconnect-loop forever)
- `watch` is one session per connection — hence a tab *is* a connection — and it's sent from the socket's open handler only, resuming from the `endOffset`/`rev` byte cursor so a reconnect replays the gap rather than the whole tail

**Renderer** is OpenTUI (`@opentui/core` + `@opentui/react`), which powers opencode's TUI, so the long-streaming-transcript path is proven at our workload.

**Structure.** `src/client/` is the server-as-types: a pure reducer over frames, injectable socket and fetch, no terminal and no React. `src/ui/` is the terminal. Keys resolve to `Action`s through a pure keymap state machine and `runAction` is the only mutator, so "what does this key do" is two files. UI position lives in a ref-backed store because a terminal delivers `^b w` as two keypresses in a single tick, before React re-renders.

## Verification

49 tests, ~2s, no server and no terminal required — including render tests that mount the real app into an in-memory terminal against a fake `fetch` and a fake WebSocket, press real keys, and assert on the resulting screen.

Those caught four real bugs while building, each of which would have shipped:

1. a merge that let a stale field win, downgrading a committed `tool_result` back to `tool_use`
2. store getters passed unbound to `useSyncExternalStore`, losing `this` at mount
3. the key-burst race above — `^b w` silently ate the second key
4. a workspace-name fetch race that painted raw `prj-…` ids on first frame

Also verified live against the running server (session list, auth, watch, transcript render, help overlay, detach) both from source and as a compiled binary, driven through a pty.

## Decisions

- **Tabs, no splits.** Splits were most of the layout complexity for a layout nobody had asked for; `^b w` plus `ctrl+←/→` covers the same ground.
- `ctrl+enter` (steer) is undistinguishable from a bare enter without the kitty keyboard protocol, so `os` requests it at startup and `alt+enter` is the portable alias.
- `os1-tui` is excluded from the root `tsconfig.json`: its JSX resolves to OpenTUI's intrinsic elements, not React DOM's. Check it with `cd os1-tui && bun run typecheck`.

## Still open

Phase 4 — diff pane, PR panel, and a terminal pane over the existing `term_start` (as a full-screen alt-screen handoff; a live pty inside a TUI pane needs an emulator we don't have). Plus published per-target binaries and an `install.sh` client-only mode, and mouse support.

https://claude.ai/code/session_014SRVCZ2EAF4cge8JPHNWgK

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -66,9 +66,47 @@ jobs:
       #
       # Fixing this means finding the cross-file state (this codebase parks a lot
       # on globalThis) and isolating it. Then delete continue-on-error.
+      # Scoped to the server suite on purpose: `os1-tui/` is a separate package
+      # with its own node_modules, so the root sweep can't resolve its imports —
+      # and it doesn't need to, because the os1-tui job below gates it properly.
+      # These two directories are where every server-side test lives.
       - name: Unit tests (reported, not yet a gate — see comment)
         continue-on-error: true
+        run: bun test src scripts
+
+  # The terminal client is its own package (own deps, own tsconfig — its JSX
+  # resolves to OpenTUI's elements, not React DOM's), so it gets its own job.
+  # Unlike the server sweep above this one IS a gate: the suite is isolated,
+  # deterministic, and needs neither a server nor a terminal.
+  os1-tui:
+    name: os1-tui (terminal client)
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v4
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6
+        with:
+          bun-version: 1.3.14
+
+      - name: Install dependencies
+        working-directory: os1-tui
+        run: bun install --frozen-lockfile
+
+      - name: Type-check
+        working-directory: os1-tui
+        run: bun run typecheck
+
+      - name: Tests
+        working-directory: os1-tui
         run: bun test
+
+      # The whole point of the package is that it compiles to a binary you can
+      # drop on a laptop; a build break there is invisible to the tests.
+      - name: Compiles to a standalone binary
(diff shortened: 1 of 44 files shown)


tella-butler:
Superseded by #80. This branch accidentally carried seven unrelated commits — I branched in the shared checkout, so other sessions' commits stacked on top of mine. #80 is the same work rebuilt cleanly on master (and caught a missing `@types/react` in the process). Nothing is lost: this exact history is preserved on `rescue/os1-tui-mixed`.

PR #85, opened 2026-08-03
Borrow qm's harness ideas: silence declarations, a command-policy floor, and a judged memory bench

tella-butler:
Three self-contained ports from [yc-software/qm](https://github.com/yc-software/qm) (MIT), analysed against OpenSession earlier today. One commit per feature.

## 1. `opensession-turn`: unattended runs declare their silence

An unattended run that ends without reaching anyone is ambiguous — "correctly quiet" and "lost the thread" look identical (the Plain-triage failure: spawn a sub-agent, end on "ok", no note posted). New in-process MCP server carries `finish_silently` / `stay_silent`; `runAgent` watches checked kinds (automation / plain / action / security-scan / github-*) for outward-effect tool calls and settles a verdict at run end: **reached** / **declared** / **silent-drop**. A silent-drop logs a papercut + audit event, never fails the run. Ledger keyed strictly on the bks session id so a run is never judged on a ledger it couldn't write into.

## 2. Org-floor command policy for unattended bash

`deniedTools` is name-level; nothing stopped an automation run steered by untrusted ticket text from `rm -rf`, force-pushing, or `curl | sh`. Ported qm's `scannableCommand` normalizer (written-heredoc stripping, quote unwrapping with bare-word preservation, ANSI-C decoding, `$(…)`/backtick extraction, recursion into `bash -c`/`eval`/`env -S`/`xargs`/wrapper chains/literal-stdin-to-shell/here-strings/variable indirection, depth 8) plus its evasion test corpus wholesale.

Enforcement: unattended code-mode runs get `bash: {"*": "ask"}` and both permission bridges answer each ask via `bashAskPolicyReply` — deny rejects everywhere, require_approval rejects unattended (instructions tell the agent to propose the command in its note), allowed commands answer their own ask. **Interactive runs are untouched** (kind-based gating, not `policy.unattended`, so the slack/linear loops gain no wedge surface). Floor = qm's five rules + `git reset --hard`/`git checkout .` (shared-checkout protection). Framed honestly as a tripwire for confused/nudged agents, not a boundary against an adversarial model.

## 3. `bun run bench:memory`: judged memory bench

Canned conversations (qm's six fixtures + one Michael-flavored ops-channel one) replayed through three strategies — `agent-only` (current behavior, simulated as REMEMBER/FORGET ops), `per-turn` extraction, `consolidated` (per-turn + one UPDATE/DELETE/ADD pass) — each notebook scored by an LLM judge on signalToNoise / staleness / inferenceVsObservation. Pure logic in `src/server/memory-bench.ts` (unit-tested, no model); model calls only in the script.

First live smoke on the supersession fixture: per-turn scores **staleness 2/10** — stale facts accumulate without curation, which is exactly the case for adopting consolidation, and now it's a number instead of an argument. Full 3-strategy baseline is running; numbers to follow as a comment.

⚠️ Side finding from the smoke: the judge flagged the extractor attributing facts to "Michiel" — the shared one-shot server appears to leak global `~/.claude/CLAUDE.md` context into utility calls. Pre-existing for all one-shots (titles, classifiers), not touched here; worth its own look.

## Notes

- `bun test src/` green (1282 pass), `tsc --noEmit` clean.
- Runner/bridge changes ⇒ needs a real `systemctl restart opensession` after merge.
- Not taken from qm (deliberately): the scope/ACL lattice, Postgres-everything, zero-comments style — per the analysis, the vocabulary is worth more than the machinery at our size.

Started by Michiel in this Michael session

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   },
   "scripts": {
     "app:dev": "bun scripts/app-dev.ts",
+    "bench:memory": "bun scripts/memory-bench.ts",
     "dev": "bun --hot opensession.ts",
     "opensession": "bun scripts/cli.ts",
     "setup": "bun scripts/cli.ts onboard",
--- a/scripts/memory-bench.ts
+++ b/scripts/memory-bench.ts
@@ -1,0 +1,112 @@
+/**
+ * Memory-strategy benchmark runner — `bun run bench:memory`.
+ *
+ * Replays the canned conversations in test/memory-bench/*.json through each
+ * strategy in src/server/memory-bench.ts and has an LLM judge score every
+ * resulting notebook. Real model calls on the opencode engine (the same
+ * opencodeOneShot path titles and classifiers use), so this is a SCRIPT, run
+ * deliberately — never part of bun test.
+ *
+ * Env:
+ *   MEMORY_BENCH_STRATEGIES  comma list to run (default: all)
+ *   MEMORY_BENCH_FIXTURES    comma list of conversation ids (default: all)
+ *   MEMORY_BENCH_MODEL       extraction/curation model (default: oneshot default, haiku)
+ *   MEMORY_BENCH_JUDGE_MODEL judge model (default: claude-sonnet-5 — judge quality
+ *                            bounds the whole bench, don't skimp it)
+ *   MEMORY_BENCH_OUT         path for a JSON report of rows + full notebooks
+ *
+ * Read the notebooks in the report, not just the table — a judge can be
+ * fooled, and the per-conversation notebooks are small enough to eyeball.
+ */
+import { readdirSync, readFileSync, writeFileSync } from "fs";
+import { join } from "path";
+import { opencodeOneShot } from "../src/server/opencode-oneshot";
+import {
+  BENCH_STRATEGIES,
+  formatTable,
+  JUDGE_PROMPT,
+  parseBenchConversation,
+  parseJudgeVerdict,
+  renderJudgeInput,
+  renderNotebook,
+  summarize,
+  type BenchResult,
+  type OneShot,
+} from "../src/server/memory-bench";
+
+const requested = (process.env.MEMORY_BENCH_STRATEGIES ?? "")
+  .split(",")
+  .map((s) => s.trim())
+  .filter(Boolean);
+const strategies = requested.length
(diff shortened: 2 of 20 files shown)


tella-butler:
## Memory bench baseline — the result inverts what I expected

Full run, 7 fixtures × 3 strategies, judged on sonnet:

| strategy | convs | signal/noise | staleness | infer-vs-obs | **overall** |
|---|---|---|---|---|---|
| **agent-only** (what we ship) | 7 | 8.4 | 9.3 | 7.6 | **8.4** |
| consolidated | 7 | 7.6 | 8.6 | 6.7 | **7.6** |
| per-turn (qm's default) | 7 | 4.1 | 5.3 | 4.4 | **4.6** |

**Our current agent-curated memory wins, clearly.** Per-turn extraction — the strategy I went in expecting to adopt — is far worse on every axis, and consolidation rescues most of that gap (4.6 → 7.6) without closing it.

The per-fixture detail explains why. Per-turn scores `staleness 0` on stale-fact-supersession (it appends "US Eastern" and "US Pacific" as coexisting facts and never supersedes) and `signalToNoise 0` on preferences-and-identity. An agent deciding what to keep, with the notebook in front of it, beats an extractor that sees one exchange at a time — which is exactly the case for the tool-driven design we already have.

So: **don't adopt per-turn extraction.** The bench's first job turned out to be stopping a change, which is the cheapest kind of value it can deliver.

Two follow-ups it does suggest, neither urgent:
- **Consolidation as maintenance, not as a strategy.** It clearly repairs an accumulating notebook. Worth trying as a periodic pass over our existing stores rather than as a replacement for curation — the bench can measure that directly.
- The weakest axis for agent-only is `inferenceVsObservation` (7.6), dragged down by noise-heavy-debugging (5) and ops-channel-noise (5) — both the "lots of operational chatter" shape. If we tighten the `store_memory` wording, that's the number to watch.

Caveat on reading these: single judged run, no repeats, so treat ±1 as noise. Notebooks are in the JSON report and are short enough to eyeball — worth doing before acting on any of this.

Also confirms the side finding from the smoke: per-turn notebooks attribute facts to "Michiel" on conversations that never name the user, which points at global `~/.claude/CLAUDE.md` context leaking into `opencodeOneShot` calls. Pre-existing for every one-shot (titles, classifiers, intent) — worth its own look, unrelated to this PR.

PR #94, opened 2026-08-12
Stop native transcript pagination jumps

happylinks:
## Summary
- wait for scroll deceleration to finish before automatically loading earlier transcript history
- preserve the request-time viewport and restore it once, only after prepended content has genuinely grown
- prevent lazy-loader reappearance from paging repeatedly during one gesture

## Verification
- iOS simulator scheme build
- macOS scheme build
- macOS test suite
- instrumented 5,901-entry session replay on iPhone 17 Pro simulator

Created by [this Michael session](https://os.tella.dev/session/os-019ff503-1bd9-7000-8e72-36b5482f565f)

### DIFF
--- a/os1-ios/OS1/Views/SessionView.swift
+++ b/os1-ios/OS1/Views/SessionView.swift
@@ -38,41 +38,45 @@ struct SessionView: View {
         horizontalSizeClass == .compact ? 16 : 20
     }
 
-    /// Where the reader was when a page of earlier history landed, measured
+    /// Where the reader was when a page of earlier history was requested, measured
     /// from the END of the transcript — the one coordinate the page doesn't
-    /// move (see `TranscriptScroll.distanceFromEnd`). Set from the scroll
-    /// geometry's PREVIOUS value the first time the content grows, which is
-    /// the last thing measured before the page was laid out.
+    /// move (see `TranscriptScroll.distanceFromEnd`). Capturing before the
+    /// request avoids depending on whether SwiftUI reports the new geometry
+    /// before or after the view model publishes the page.
     @State private var prependDistanceFromEnd: CGFloat?
     /// The content height before the page landed. A lazy stack can briefly
     /// report less than this while re-realizing its rows; that is not a valid
     /// restoration target yet.
     @State private var prependBaselineContentHeight: CGFloat?
-    /// Armed when a page lands, disarmed by the height change it is waiting
-    /// for.
+    /// Armed when a page lands, disarmed by its one position correction.
     @State private var awaitingPrepend = false
     @State private var prependRestoreTask: Task<Void, Never>?
     /// A page requested before the reader began a newer gesture must not claim
     /// their position when its delayed response finally arrives.
     @State private var prependRequestInteraction: Int?
     @State private var scrollInteractionGeneration = 0
+    /// One automatic page per scroll gesture. A prepend can make the lazy
+    /// loader disappear and reappear while rows settle; that is layout work,
+    /// not another request from the reader.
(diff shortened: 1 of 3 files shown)


tella-butler:
<details>

### 🤖 Michael review

🔄 Reviewing `ad0e2ba`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-94-review)

</details>

tella-butler:
Safe once the P1 below is fixed. The PR improves pagination timing and prevents repeated automatic loads during one gesture, but the restoration can still disarm before the prepended page is laid out. Jump-to-start also loses its previous one-shot suppression when the reader interacts during the walk.
_2 inline comments below._
<sub>Reviewed `a99c49d` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-94-review)</sub>

os1-ios/OS1/Views/SessionView.swift
```diff
@@ -1119,58 +1130,45 @@ struct SessionView: View {
...
-                if height == lastRestoredHeight {
-                    unchangedHeightRetries += 1
-                    guard unchangedHeightRetries <= 2 else { continue }
-                } else {
-                    lastRestoredHeight = height
-                    unchangedHeightRetries = 0
-                }
-                scrollPosition.scrollTo(y: y)
+            for _ in 0..<30 {
+                restoreAfterPrependIfPossible(scrollGeometry.value)
```
tella-butler, line 1148: 🔴 **P1** — Request-time height does not prove the prepended page was laid out

`beginPrependRestore` immediately tests the cached geometry against the height captured when the request started. Existing rows are documented to keep growing asynchronously, so that comparison can succeed before the requested page changes layout. For example, with the measured 11,495pt baseline, an existing markdown row can settle to 11,520pt while the response is pending. When `historyPrependSeq` fires, this first poll accepts 11,520 because it is greater than 11,495, clears `awaitingPrepend`, and scrolls before the new rows are measured. When the page then raises the content to 12,603pt, the geometry callback is ignored and the reader can still jump roughly 1,083pt. Keep the request-time distance, but only disarm after a geometry transition attributable to the landed page, rather than treating any growth since the request as proof of the prepend.

tella-butler (commented):
Michael review · `a99c49d`

os1-ios/OS1/Views/SessionView.swift
```diff
@@ -865,27 +864,39 @@ struct SessionView: View {
...
+            prependBaselineContentHeight = geometry.contentHeight
+        }
         prependRequestInteraction = scrollInteractionGeneration
         viewModel.loadEarlier()
     }
 
     private func requestJumpToStart() {
         guard viewModel.canLoadEarlier, !viewModel.loadingEarlier else { return }
+        autoEarlierPendingIdle = false
+        autoEarlierRequestInteraction = scrollInteractionGeneration
```
tella-butler, line 899: 🟠 **P2** — Jump completion can automatically page past its ceiling

The automatic-request generation is recorded when the jump starts, but a jump can take several page requests and the transcript remains interactive throughout. If the reader touches the transcript during a walk that reaches `jumpMaxEntries`, `scrollInteractionGeneration` advances while `autoEarlierRequestInteraction` remains stale. The final `scrollTo(first, anchor: .top)` realizes the loader, whose `onAppear` then sees a new generation and automatically requests another page. This defeats the documented ceiling behavior that should leave further loading deliberate, and can mutate the transcript immediately after landing. Re-arm the automatic-request marker with the current generation in the `jumpLandedSeq` handler before scrolling, or retain an independent one-shot suppression for the loader exposed by jump completion.

PR #95, opened 2026-08-13
Stop the review canvas scrolling sideways on a phone

happylinks:
## Why

On a phone the `/workspace/<id>/review` Conversation tab put the whole review surface into horizontal scroll — the sticky sub-tab row, the PR title and the comment bodies all slid off-screen together. Measured over CDP at 390x844: `main` scrollWidth **930** against a 390 client width.

## What

Three causes, each measured rather than guessed:

- **The conversation column never shrank.** It's a flex item with `mx-auto`; an auto cross-axis margin opts an item out of stretching, so the box sized to its content and `max-w-[760px]` became a *fixed* 760px width — wider than the phone before a single comment rendered. `w-full` restores stretch and leaves `max-w` as the cap it reads as.
- **Comment cards and check sections are grid items**, whose automatic minimum size is their min-content width. One Vercel deploy table was enough to widen the track to 916px and drag the scroller with it. `min-w-0` lets them shrink; the table keeps its own `overflow-x`.
- **`.markdown` wrapped at `normal`**, so a module path or a bare URL both overflowed its line and fed the min-content width driving the above. `overflow-wrap: anywhere` fixes both. `pre` is unaffected — it never wraps and keeps its horizontal scroll. (This subsumes the same declaration that `.review-preview-markdown` already carried, which is removed.)

Also stacks the bottom review bar on phones: `Open workspace` / `Close` / `Finish review` are together wider than the bar, so they overflowed its right edge and squashed the pending-comment caption to roughly one word per line.

## Verification

CDP at 390x844 against a real PR. `main` scrollWidth vs clientWidth:

| tab | before | after |
| --- | --- | --- |
| Conversation | 930 / 390 | 390 / 390 |
| Commits | 390 / 390 | 390 / 390 |
| Checks | 397 / 390 | 390 / 390 |
| Files changed | 390 / 390 | 390 / 390 |

Desktop 1440x900 re-shot and unchanged; the session transcript (which shares `.markdown`) re-shot at phone width with no regression. `bun test src/frontend` and `tsc --noEmit` show the same pre-existing failures as `main`, nothing new.

## Not covered

Web UI only. `os1-ios/` renders its own transcript and PR surfaces, so nothing here reaches the native app; I didn't find an equivalent overflow there, but I also didn't audit it.

### DIFF
--- a/src/frontend/components/PrPanel.tsx
+++ b/src/frontend/components/PrPanel.tsx
@@ -1228,7 +1228,7 @@ export function PrPanel({
             inside the one scroll container so the identity scrolls away with
             the diff. Only the tab row sticks, so the reviewer keeps a way back
             to Conversation/Commits/Checks once they're deep in a file. */}
-        <main className="min-h-0 flex-1 overflow-y-auto bg-surface pb-24 [--review-file-header-top:106px]">
+        <main className="min-h-0 flex-1 overflow-y-auto bg-surface pb-24 [--review-file-header-top:106px] phone:pb-36">
           <header className="flex min-h-[96px] shrink-0 items-center gap-5 px-6 py-4 phone:min-h-[78px] phone:px-3">
             <div className="min-w-0 flex-1">
               <a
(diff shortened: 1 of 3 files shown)


tella-butler:
Safe to merge. The changes address each verified source of mobile horizontal overflow while preserving intentional scrolling for tables and preformatted blocks. The responsive review bar also reserves sufficient phone-sized bottom space, and no affected shared markdown consumer exposes a conflicting layout contract.
<sub>Reviewed `51f2b3d` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-95-review)</sub>

tella-butler:
Before/after on the page that was reported — `/workspace/ws-a125caab…/review` → Conversation, PR #5602, CDP at 390x844 CSS, DPR 3.

**Before** (`main`): `main` scrollWidth **925** against a 390 client width. Comment text runs off the right edge — `/api/streamin…`, `Selectors__Zoom.eligibleF…`, `su_cmslxag…` are all cut, and the whole surface (title, sticky sub-tab row) scrolls with it.

Before: conversation tab text running off the right edge at 390px

**After** (this branch): **390 / 390**. Same PR, same viewport.

After: conversation tab wrapping inside 390px

happylinks (commented):
Ignore: diagnostic probe from an Open Session debugging session (investigating a 422 on the reviews API). No action needed. Submitted reviews cannot be deleted, so this note replaces the original text.

PR #97, opened 2026-08-16
Capture UI: drive the page before the screenshot

tella-butler:
`scripts/capture-ui.ts` can navigate and screenshot, but there was no hook to drive the page in between. Six sessions each hand-copied 100 to 150 lines of it into /tmp in a single day, always for the same reason: to add a click, a hover, or to undo the freeze stylesheet. 15 of 64 agent-logged papercuts on 2026-08-15 name this one script.

### What is new

**Page-driving steps, repeatable and ordered.** `--click <selector>`, `--hover <selector>`, `--eval <js>`, `--script <file.js>` run after navigation and the emulation check, in the order written, so `--hover x --click y` is a sequence. Clicks and hovers go through `Input.dispatchMouseEvent` because a synthetic `.click()` does not navigate this app. `--eval` and `--script` run inside an async wrapper, so a step can `await`, and a `return` prints the value. `--step-wait` (default 500ms) paces them; the existing settle loop still runs before the capture.

Selector resolution takes the first match with a non-zero box, not the first match. Several controls render twice here (a phone bar and a desktop rail) and the copy the reader cannot see is regularly first in document order: `[aria-label="Open settings"]` matches 2 elements on `/`, and the off-screen one is first.

**`--motion`** skips the injected freeze stylesheet and emulates `prefers-reduced-motion: no-preference` instead of `reduce`. It still hides the caret, which is neither a transition nor an animation and otherwise blinks into the frame at random. **`--seek <ms>`** pauses `document.getAnimations()` at a fixed point and implies `--motion`, since seeking is meaningless once the freeze has set `animation: none`. With live motion and no seek the two-identical-frames settle loop can never converge, so that combination takes a single shot.

**The emulation assert names the failing field.** It printed all six values when one tripped, which reads as a viewport problem; one session lost a round-trip to it. Now: `capture emulation did not apply: width is 1441, expected 1440 (probe {...})`.

**The CDP preamble moved to `scripts/lib/cdp-session.ts`.** Two papercuts asked for this directly. `openCdpSession()` leases the bounded browser, opens a target, connects the socket and returns `{ send, evaluate, close }`, so a genuine one-off no longer has to copy 60 lines to get a `send()`. It also rejects on a CDP `message.error` rather than resolving `undefined`, which is what let a failed command surface as whatever asserted the page state next. No other script was touched.

Argument parsing now tracks which values belong to a flag, so `--click button out.png` no longer reads `button` as the output path.

### Deviations from the brief

- The brief said `tellahq/backstage`; the remote is `tellahq/opensession` (renamed), so the PR is there.
- The brief offered the `cdp-session` extraction as optional. I did it, because it is the other half of the same papercut and it is what makes `send()` reject on error.
- No walkthrough: this is a developer script with no user-visible surface.

### Verified

`bunx tsc --noEmit` in a detached worktree at `615160d6`: **0 errors before, 0 errors after** (baseline measured on the same clean tree, not assumed).

Every command below was run from the worktree against the live server. Default mode is unchanged when no new flag is passed.

```
# 1. default, unchanged
bun scripts/capture-ui.ts /tmp/capture-hooks-shots/default.png --route /
#   -> CSS viewport 1440x900; PNG 2880x1800; DPR 2; Electron material

# 2. click + eval (proves the click navigated, and picked the visible duplicate)
bun scripts/capture-ui.ts /tmp/capture-hooks-shots/clicked.png --route / \
  --click '[aria-label="Open settings"]' --eval 'return location.pathname'
#   -> --eval returned "/settings"
#   -> ...; Electron material; 2 steps

# 3. motion + seek
bun scripts/capture-ui.ts /tmp/capture-hooks-shots/seek.png --route /settings \
  --wait 400 --motion --seek 120
#   -> paused 12 animations at 120ms
#   -> ...; motion at 120ms

# 4. hover + script file (statements, await, return)
bun scripts/capture-ui.ts /tmp/capture-hooks-shots/hovered.png --route / \
  --hover '[aria-label="Open settings"]' --script /tmp/capture-hooks-step.js
#   -> --script returned 2

# 5. a selector that matches nothing fails loudly
bun scripts/capture-ui.ts /tmp/x.png --route / --click '.does-not-exist'
#   -> error: no element with a visible box matched .does-not-exist (0 matched the selector)

# 6. emulation assert, forced with a throwaway copy patched to report innerWidth + 1
#   -> error: capture emulation did not apply: width is 1441, expected 1440 (probe {...})
```

All four PNGs are 2880x1800 with 9k to 14k distinct colours (not a flat frame from a dead app), and `default.png` vs `clicked.png` differ in 2,104,561 pixels, i.e. home versus settings.

Started by Dreaming (nightly reflection) in [this Michael session](https://os.tella.dev/session/os-01a008db-577a-7000-bef9-0cbd04043bf3)

### DIFF
--- a/scripts/capture-ui.ts
+++ b/scripts/capture-ui.ts
@@ -1,24 +1,68 @@
 #!/usr/bin/env bun
 /** Capture the running app with Mac/Retina-quality desktop defaults. */
-import { mkdirSync, writeFileSync } from "node:fs";
+import { mkdirSync, readFileSync, writeFileSync } from "node:fs";
 import { dirname, resolve } from "node:path";
-import {
-  acquireCdpBrowser,
-  closeCdpTarget,
-  releaseCdpBrowser,
-} from "./lib/cdp-browser";
+import { openCdpSession } from "./lib/cdp-session";
 import { localAutomationToken } from "./lib/local-auth";
 import { captureInitScript, captureViewport } from "./lib/visual-capture";
 
 const argv = process.argv.slice(2);
-const outputArg = argv.find((value) => !value.startsWith("--"));
+
+// Flags that consume the next argument. Tracking them is what keeps a selector
+// or a route from being read as the output path.
+const VALUE_FLAGS = new Set([
+  "route",
+  "width",
+  "height",
+  "theme",
+  "wait",
+  "step-wait",
+  "seek",
+  "click",
+  "hover",
+  "eval",
+  "script",
+]);
+type StepKind = "click" | "hover" | "eval" | "script";
+const STEP_FLAGS = new Set<string>(["click", "hover", "eval", "script"]);
+
+// Steps run in the order they were written, so --hover x --click y is a
+// sequence rather than a set.
+const steps: { kind: StepKind; value: string }[] = [];
+const consumed = new Set<number>();
+for (let index = 0; index < argv.length; index++) {
+  const arg = argv[index]!;
+  if (!arg.startsWith("--")) continue;
+  const name = arg.slice(2);
+  if (!VALUE_FLAGS.has(name)) continue;
+  const value = argv[index + 1];
+  if (value === undefined) throw new Error(`--${name} needs a value`);
+  consumed.add(index + 1);
+  if (STEP_FLAGS.has(name)) steps.push({ kind: name as StepKind, value });
+}
+
+const outputArg = argv.find(
+  (value, index) => !value.startsWith("--") && !consumed.has(index),
+);
 const flag = (name: string, fallback?: string) => {
   const index = argv.indexOf(`--${name}`);
   return index >= 0 ? argv[index + 1] : fallback;
 };
 if (!outputArg) {
   console.error(
-    "usage: bun scripts/capture-ui.ts <output.png> [--route /] [--width 1440] [--height 900] [--theme light|dark] [--web] [--wait 3000]",
+    [
+      "usage: bun scripts/capture-ui.ts <output.png> [--route /] [--width 1440] [--height 900]",
+      "         [--theme light|dark] [--web] [--wait 3000]",
+      "  drive the page before capturing (repeatable, runs in the order written):",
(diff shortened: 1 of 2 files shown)


tella-butler:
Safe once the P1 below is fixed. The ordered interactions and CDP extraction fit the existing bounded-browser design, but the default timing causes the new motion controls to miss the finite transitions they are meant to capture.
_1 inline comment below._
<sub>Reviewed `c2827b9` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-97-review)</sub>

scripts/capture-ui.ts
```diff
@@ -124,36 +223,79 @@ try {
...
+        .join("; ")} (probe ${JSON.stringify(actual)})`,
     );
   }
 
-  let previous = "";
+  for (const step of steps) {
+    if (step.kind === "hover") await movePointer(step.value);
+    else if (step.kind === "click") await clickAt(step.value);
+    else await runSource(step.kind, step.value);
+    await sleep(stepWaitMs);
```
tella-butler, line 255: 🔴 **P1** — Seek runs after the app's transitions have finished

Every step sleeps for 500ms by default before `document.getAnimations()` is queried. This app's verified transition durations are 150ms, 200ms, and 280ms; for example, opening Settings starts `settings-paper-in` for `--dur` (200ms), while the phone settings pager uses `--dur-lg` (280ms). Therefore `--click '[aria-label="Open settings"]' --seek 100` waits until that animation is finished and no longer returned by `document.getAnimations()`, then captures the final state instead of the requested 100ms frame. `--motion` without seek has the same problem for finite motion. Preserve the pacing between earlier steps, but after the final motion-triggering step wait only until the animation has been created, such as the next frame, then pause/seek or capture before applying the normal step wait.

tella-butler (commented):
Michael review · `c2827b9`

PR #98, opened 2026-08-18
Scheduler: catch up cron minutes a tick never landed in

tella-butler:
## The bug

`startScheduler` runs a 20s tick and evaluates only the minute that tick happened to wake up in. A minute that no tick lands in skips every automation scheduled for it until that cron's next occurrence, which for a daily is 24 hours later. There was no catch-up.

Measured on 2026-08-17: a self-restart at 14:01:46 left a 125 second gap in the audit log, 13:59:37 to 14:01:42. The 14:00 minute was never evaluated, so both hourly automations (`0 * * * *`, "Michael Health Monitor" and "Cassandra") recorded 23 of 24 hours that day. "Production Watchdog" at `20 * * * *` was unaffected because its minute came later. Same shape on 2026-08-14. Four sightings total.

## The fix

`pendingMinutes(lastMinute, now, max)` in `cron.ts` returns every whole minute after the one the scheduler last recorded, up to and including the current minute, capped at `MAX_CATCHUP_MINUTES = 5`, oldest first. The tick fires when any of them matches:

```ts
if (minutes.some((minute) => cronMatches(schedule, minute)))
```

`some`, not a loop, so one tick is at most one run per automation: a caught-up gap spanning several matching minutes (a `* * * * *` schedule) never fires a burst. Consecutive ticks own disjoint minute ranges, so no minute is evaluated twice and a slot cannot double-fire.

`runOnceAt` still fires against `now`, never against a caught-up minute.

## A fresh boot catches up nothing

This is the part worth reviewing. On a fresh scheduler (`lastFiredMinute === ""`) `pendingMinutes` returns the current minute only. Replaying the last 5 minutes on boot would fire a burst of backdated automations on every restart, which is a worse bug than the miss it papers over. The current minute is still evaluated exactly as it is today, so booting early in a minute does not itself skip that minute's slot. Both behaviours are pinned by tests.

**Honest limitation:** the measured incidents were all process restarts, and a restart resets `lastFiredMinute`, so this change does **not** recover the 14:00 slot in the 2026-08-17 sighting. That case is excluded by the no-replay-on-boot rule above, deliberately. What this fixes is every other cause of a missing tick within a live process: event-loop stalls, a suspended or overloaded host, and tick drift, where a 20s interval and a 60s window leave no margin. Closing the restart case properly needs a durable record of the last evaluated minute plus a policy for how stale a slot may be before it is dropped, which is a larger change than this one.

## Tests

Seven cases over `pendingMinutes` and six over the scheduler's tick decision, in `cron.test.ts` (`bun test src/server/cron.test.ts`: 46 pass, 0 fail):

- a 2 to 3 minute gap fires a `0 * * * *` automation whose minute fell in the gap
- a fresh boot fires nothing retroactively, and still fires the boot minute's own slot
- a gap longer than `MAX_CATCHUP_MINUTES` is clamped to the newest 5 and a `* * * * *` schedule fires once, not 30 times
- no double-fire: a slot covered by both catch-up and the following ticks runs exactly once
- an unparseable stored minute falls back to the current minute rather than replaying the window (a shape check, because `Date.parse` is lenient enough to turn a garbled key into a finite very old timestamp; caught by the test)

## Verification

- `bunx tsc --noEmit`: byte-identical to the baseline at the base commit, 0 errors both sides.
- `bun test src/server`: failure lists diffed against a detached worktree at the base commit. Baseline fluctuates between 31 and 32 failures run to run; the 32-failure baseline list is **identical** to this branch's. The mover is `report signals > persists urgency, confidence, and bounded structured highlights`, which fails in the unmodified baseline too and passes when its file is run alone. `reports.test.ts` writes into the live `~/.opensession-reports/`, shared with the running server and other sessions' test runs. No new failures from this change.

Started by Dreaming (nightly reflection) in [this OS session](https://os.tella.dev/session/os-01a01327-fe28-7000-b9f8-f5138044a420)

### DIFF
--- a/src/server/automations.ts
+++ b/src/server/automations.ts
@@ -13,7 +13,7 @@ import {
   webhookBodyTooLargeResponse,
 } from "./shared/bounded-body";
 import { labelIdentity } from "./shared/user-mappings";
-import { parseCron, cronMatches, nextRun } from "./cron";
+import { parseCron, cronMatches, nextRun, minuteKey, pendingMinutes } from "./cron";
 import {
   STRIPE_CONFIRM_TOOLS,
   declaredRunFailure,
@@ -1650,9 +1650,15 @@ export function startScheduler(onSessionCreated?: (sessionId: string) => void): 
 
   schedulerInterval = setInterval(() => {
     const now = new Date();
-    const minuteKey = now.toISOString().slice(0, 16);
-    if (minuteKey === lastFiredMinute) return;
-    lastFiredMinute = minuteKey;
+    const currentMinute = minuteKey(now);
+    if (currentMinute === lastFiredMinute) return;
+
+    // Every whole minute this tick is responsible for, not just the one it woke
+    // up in: a minute that no tick lands in would otherwise skip its automations
+    // until the cron's next occurrence, 24 hours later for a daily. Bounded by
+    // MAX_CATCHUP_MINUTES, and a fresh scheduler catches up nothing.
+    const minutes = pendingMinutes(lastFiredMinute, now);
+    lastFiredMinute = currentMinute;
 
     for (const automation of listAutomations()) {
       if (!automation.enabled) continue;
@@ -1671,8 +1677,13 @@ export function startScheduler(onSessionCreated?: (sessionId: string) => void): 
         continue;
       }
 
-      if (!automation.schedule) continue;
-      if (cronMatches(automation.schedule, now)) {
+      const schedule = automation.schedule;
+      if (!schedule) continue;
+      // `some`, not a loop over `minutes`: one tick is at most one run per
+      // automation, so a caught-up gap that spans several matching minutes
+      // (a "* * * * *" schedule, say) never fires a burst. Consecutive ticks
+      // own disjoint minute ranges, so no minute is ever evaluated twice.
+      if (minutes.some((minute) => cronMatches(schedule, minute))) {
         // Fire and forget — runner guards against overlap per automation
         void runAutomation(automation, onSessionCreated, { trigger: "cron" });
       }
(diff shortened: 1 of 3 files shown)


tella-butler:
Safe once the P2 below is fixed. The bounded catch-up window handles forward gaps correctly, but the scheduler can still evaluate the same cron slot twice after the system clock moves backward.
_1 inline comment below._
<sub>Reviewed `26bb66f` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-98-review)</sub>

tella-butler (commented):
OS review · `26bb66f`

src/server/automations.ts
```diff
@@ -1650,9 +1650,15 @@ export function startScheduler(onSessionCreated?: (sessionId: string) => void):
...
-    lastFiredMinute = minuteKey;
+    const currentMinute = minuteKey(now);
+    if (currentMinute === lastFiredMinute) return;
+
+    // Every whole minute this tick is responsible for, not just the one it woke
+    // up in: a minute that no tick lands in would otherwise skip its automations
+    // until the cron's next occurrence, 24 hours later for a daily. Bounded by
+    // MAX_CATCHUP_MINUTES, and a fresh scheduler catches up nothing.
+    const minutes = pendingMinutes(lastFiredMinute, now);
+    lastFiredMinute = currentMinute;
```
tella-butler, line 1661: 🟠 **P2** — Do not move the minute watermark backward

`pendingMinutes` deliberately returns `[]` when `lastFiredMinute` is ahead of the current clock, but this assignment then replaces that future watermark with the earlier minute. For example, after evaluating through `14:05`, a clock correction to `14:02` sets the watermark to `14:02`; when time reaches `14:03` again, a `3 * * * *` automation fires a second time for the same slot. This contradicts the no-double-fire guarantee and can repeat automation side effects. Keep the existing watermark whenever there are no pending minutes.

```suggestion
if (minutes.length === 0) return;
    lastFiredMinute = currentMinute;
```

PR #99, opened 2026-08-18
Protect personal MCP grants and proxy HTTP OAuth server-side

soutar:
## Goal

Make personal MCP connections materially safer without adding a root requirement or breaking the default single-user install.

## What changed

- Personal grant storage migrates from plaintext to authenticated AES-256-GCM ciphertext on first read.
- Operator installs can supply a 32-byte `mcp-oauth-key` systemd credential. Rootless simple-mode installs mint a `0600` key beside the store, so installation and first run need no privileged step.
- Personal grants for remote HTTP MCP servers run through coordinator-side in-process proxies. Provider tokens do not enter engine config, process environments, arguments, projected sandbox files, or transcripts.
- HTTP grants are bound to their configured upstream URL. Repointing a server name fails closed, including during legacy-store migration.
- Personal tokens are no longer injected into stdio MCP executables. Those executables keep their configured workspace credential; provider-specific personal grants such as Slack can still power coordinator-owned UI actions.
- Per-user grants and `allowedUsers` visibility follow the verified prompter, not the session creator. OAuth callbacks are tied to the verified account that started them when web sign-in is enabled.
- Removing an MCP server revokes its stored registration and grants. Refreshes are single-flight and cannot resurrect a disconnected or replaced grant.

## Simple-mode behavior

The default personal install remains rootless and requires no additional setup. It has no web identity gate, so it creates shared grants in line with its single-user trust model. Existing configured stdio integrations, including Slack bot tooling, continue using their workspace credential.

The fallback key is intentionally a limited at-rest improvement: it protects a stray copy of the grant file, but not a whole state-directory backup or another process already running as the same Unix user. Full same-UID isolation still requires a separate-identity credential broker; `docs/security-model.md` states that boundary explicitly.

## Compatibility

The legacy loopback relay remains available for detached runs created before this update, but new runs use the coordinator proxy. Existing plaintext grants migrate atomically and are left unchanged if decryption or migration fails. `OPENSESSION_PERSONAL_MCP=0` disables reads and writes. The branch also fixes the current simple installer’s ShellCheck warning without changing its displayed `~/...` profile path.

## Verification

- `bun run typecheck`
- `bun scripts/check-module-side-effects.ts` (426 modules)
- focused MCP OAuth, Pi bridge, connections, policy, and catalog suites: 96 tests passed
- `bash -n install.sh`; CI’s ShellCheck gate now passes the changed line

Started by John Soutar in [this OS session](https://os.tella.dev/session/os-01a0397a-f063-7b21-8f55-d5f74848ab66)


tella-butler:
<details>

```mermaid
sequenceDiagram
  participant E as Agent engine
  participant C as Coordinator proxy
  participant K as Encrypted grant store
  participant S as stdio MCP child
  E->>C: Tool call via run-rpc
  C->>K: Decrypt selected grant
  C->>S: Spawn with provider token
  Note over S,K: stdio child is currently unconfined
```

</details>

src/server/mcp-oauth-proxy.ts
```diff
@@ -0,0 +1,179 @@
...
+      cfg.type === "sse"
+        ? new SSEClientTransport(new URL(cfg.url), { requestInit })
+        : new StreamableHTTPClientTransport(new URL(cfg.url), { requestInit });
+  } else {
+    const preset = oauthPresetFor(name);
+    if (!preset?.envVar || !cfg.command) {
+      throw new Error("Personal stdio connection has no protected OAuth preset");
+    }
+    transport = new StdioClientTransport({
+      command: String(cfg.command),
```
tella-butler: 🔴 **P1** — Run OAuth stdio servers inside the AppArmor profile

`StdioClientTransport` spawns this command directly from the unconfined coordinator. For a Slack call, that child receives the user's `SLACK_BOT_TOKEN` and also shares the coordinator's service mount namespace and Unix uid, so a compromised MCP package can read the systemd credential mount and encrypted store and recover every teammate's grants. The engine wrappers do not protect a child spawned here. Build the transport command through `secureAgentCommand` and pass the resulting wrapper command/arguments to `StdioClientTransport`.

src/server/opencode-detach.ts
```diff
@@ -365,6 +366,12 @@ export async function spawnDetachedOpencodeServer(opts: {
   mkdirSync(opts.logDir, { recursive: true });
   const logPath = `${opts.logDir}/${unit}.log`;
   const fd = openSync(logPath, "a");
+  const engineCommand = secureAgentCommand([
```
tella-butler: 🔴 **P1** — Do not adopt detached servers created before confinement

Only newly spawned detached servers pass through `secureAgentCommand`; `adoptDetachedOpencodeServers` accepts surviving registry records without checking `/proc/<pid>/attr/current` or a profile-version marker. On the first deployment, an old shared server can survive the coordinator restart and be reused unconfined. A model shell in that server can use the same coordinator `/proc/<pid>/{root,environ,fd}` paths this profile is intended to deny and recover the newly mounted key. Reject and stop records whose process is not already under `opensession-agent`, or explicitly retire all pre-profile scopes before mounting/using the credential.

packages/core/opensession-server/src/server/mcp-oauth.ts
```diff
@@ -56,12 +149,38 @@ interface ServerAuth {
...
+    const cfg = parsed.mcpServers?.[name];
+    if (!cfg) return undefined;
+    if (typeof cfg.url === "string") {
+      return { kind: "http", url: new URL(cfg.url).toString() };
+    }
+    if (typeof cfg.command === "string") {
+      return {
+        kind: "stdio",
+        command: cfg.command,
+        args: Array.isArray(cfg.args) ? cfg.args.map(String) : [],
```
tella-butler: 🔴 **P1** — Include stdio environment in the pinned binding

The stored stdio binding covers only `command` and `args`, while the proxy executes the mutable `cfg.env`. An agent can leave `command: "bun"` and its arguments unchanged but set `PATH` to a workspace directory containing a replacement `bun`; `mcpOauthBindingMatches` still succeeds, and the replacement receives the decrypted Slack token. Pin and compare every execution-affecting field consumed by the transport, including a canonicalized environment, before releasing a grant.

packages/core/opensession-server/src/server/interactive-mcp.ts
```diff
@@ -127,8 +129,19 @@ function papercutsServerFor(
 export function interactiveMcpServers(
 	user?: string,
 	sessionId?: string,
+	personalMcpScope?: McpScope,
 ): Record<string, unknown> {
 	const createdBy = user || productName();
+	const session = sessionId ? findSession(sessionId) : undefined;
+	// Personal provider tools are opt-in at the run launch sites. Other callers
+	// (notably Desk voice) deliberately consume a narrower interactive facade.
+	const personalMcp = session && personalMcpScope
```
tella-butler, line 160: 🔴 **P1** — Pass the personal MCP scope from every run launcher

Personal proxies are now opt-in, but `session-create.ts:644` still calls `interactiveMcpServers(spec.user, bksId)` and `runner-session.ts:78` does the same for every Runner turn. Meanwhile `buildOpencodeMcpConfig` removes any server for which that user has a grant. Therefore a new local session's opening prompt, and every Runner prompt, gets neither the external server nor its proxy. For example, an opening request to post through a connected Slack account has no Slack tools, although a later ordinary local prompt does. Pass `spec.runMcpServers ?? "all"` and `opts.mcpServers ?? "all"` at those call sites and add coverage for both launch paths.

tella-butler (commented):
OS review · `7e69d73`

src/server/mcp-oauth-proxy.ts
soutar: Fixed in 9720bbb14. `connectUpstream` now builds the stdio command through `secureAgentCommand` and hands the wrapper's command/args to `StdioClientTransport`, so the upstream package runs under `opensession-agent` like every other model-controlled process. You are right about the blast radius: the child legitimately receives its own token, but unconfined it also held the credential mount and the encrypted store, so a compromised package recovered every teammate's grant rather than the one it was given.

src/server/opencode-detach.ts
soutar: Fixed in 9720bbb14. `adoptDetachedOpencodeServersInner` now filters the registry through `retirePreConfinementRecords`, which checks `/proc/<pid>/attr/current` and stops the scope plus removes the record for anything not already enforcing under `opensession-agent`. The check is skipped entirely when the profile is not loaded, so hosts without AppArmor adopt exactly as before. This does cost surviving turns a restart on the first deploy, which seemed right given the alternative is reusing an unconfined process at the moment the key is first mounted. Parsing is split into `procAttrConfinedByAgentProfile` and covered for enforce, complain, unconfined, a foreign profile and a name that merely starts with ours.

packages/core/opensession-server/src/server/mcp-oauth.ts
soutar: Fixed in 9720bbb14. The stdio binding now carries a canonicalized `env` (sorted entries, so key order in the config is not a change), and `mcpOauthBindingMatches` compares it. I went one step stricter than the wording: a binding written before env was pinned compares as an EMPTY env rather than as a wildcard, so a server that has since grown one fails closed and asks for a reconnect instead of silently accepting whatever the config now names. Covered by a PATH-swap case that keeps command and args identical.

packages/core/opensession-server/src/server/interactive-mcp.ts
soutar: Fixed in 9720bbb14, and this one was worse than reported. Rather than patch the two call sites, I made `personalMcpScope` a required parameter, because the failure is silent in both directions: `buildOpencodeMcpConfig` drops a granted server assuming a proxy replaces it, so a launcher that forgets the scope gets neither. Making it required had the compiler find two more launchers you had not flagged, in `opensession.ts`'s resume path. Those and the run-rpc fallback builder now share one derivation, `personalMcpScopeForSession`. Desk voice passes `undefined` explicitly to keep its narrow facade. On coverage: I did not add per-launcher tests, since the type signature now enforces this at every present and future call site, which a test enumerating today's launchers would not.

tella-butler:
<details>

Not safe to merge until the confinement boundary is closed. The four previously reported issues are addressed, but a confined agent can still escape through the user systemd manager, and the rollout exposes the credential mount to surviving unconfined processes before retiring them.
_2 inline comments below._
<sub>Reviewed `9720bbb` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-99-review)</sub>

</details>

tella-butler (commented):
OS review · `9720bbb`

deploy/apparmor/opensession-agent
```diff
@@ -0,0 +1,33 @@
+#include <tunables/global>
+
+# Model-controlled engine and shell processes inherit this profile. Open Session
+# itself stays unconfined so only the coordinator can read systemd credentials.
+profile opensession-agent flags=(attach_disconnected,mediate_deleted) {
+  file,
+  network,
+  unix,
+  dbus,
```
tella-butler: 🔴 **P0** — Block escapes through the user systemd manager

The broad `file`, `dbus`, and `unix` permissions leave the per-user systemd manager callable from a confined shell. A code-mode agent can run `systemd-run --user --pipe /bin/sh -c '...'`; a transient service is spawned by the unconfined user manager, so that process does not inherit `opensession-agent`. It can then read both `~/.opensession-mcp-oauth.json` and `/proc/<coordinator>/root/run/credentials/opensession.service/mcp-oauth-key`, bypassing the entire encryption boundary. Deny access to the user manager's D-Bus/private Unix control channels from this profile and test that a confined process cannot create a user service. Denying only the `systemd-run` executable is insufficient because `systemctl`, `busctl`, or a raw protocol client can reach the same manager.

src/server/opencode-runner.ts
```diff
@@ -2284,7 +2294,7 @@ export function adoptDetachedOpencodeServers(): Promise<number> {
 }
 
 async function adoptDetachedOpencodeServersInner(): Promise<number> {
-  const records = readDetachedRegistry();
+  const records = retirePreConfinementRecords(readDetachedRegistry());
```
tella-butler: 🔴 **P1** — Retire unconfined survivors before mounting the credential

`retirePreConfinementRecords` runs inside the newly started coordinator, after systemd has already mounted `mcp-oauth-key`. A pre-PR detached engine can poll `/proc/*/root/run/credentials/opensession.service/mcp-oauth-key` while it remains unconfined and copy the key before this boot-time adoption sweep stops it. Host preview scopes explicitly survive coordinator restarts as well, but this sweep only covers registered OpenCode scopes. Stop all surviving `opensession-oc-*` and `opensession-preview-*` scopes in the installer/deploy path before restarting the unit that first gains `LoadCredential`; retain this boot check only as defense in depth.

deploy/apparmor/opensession-agent
soutar: Fixed in 7a9fb8cbd, and you are right that this made the rest of the profile decorative. The manager's control channels are denied rather than the binaries: `/run/systemd/private`, `/run/user/[0-9]*/systemd/private`, `/run/user/[0-9]*/bus`, the abstract forms, `dbus bus=session`, and `dbus bus=system peer=(name=org.freedesktop.systemd1)`. The systemd-run/systemctl/busctl execs are denied too, but only as a cheap extra, since your point about a raw D-Bus client is the reason the socket rules are the actual fix. The profile parses clean under `apparmor_parser -Q`, and the rules are pinned in agent-runtime-security.test.ts so a later edit cannot quietly drop them. Worth stating plainly for whoever reviews the tradeoff: this genuinely removes a capability agents on this instance use today, since several documented workflows start dev servers, tunnels and browser rigs with `systemd-run --user` precisely so they outlive the session scope. Those now have to go through coordinator-side tooling, which stays unconfined by design.

src/server/opencode-runner.ts
soutar: Fixed in 7a9fb8cbd. The retirement moved into the deploy path, ahead of the restart, so it happens before the unit that holds LoadCredential ever starts: it stops surviving `opensession-oc-*` and `opensession-preview-*` scopes, and the in-process sweep stays as the defence-in-depth layer you suggested. One deliberate narrowing: it runs only on the first start that GAINS LoadCredential (captured by reading the deployed unit before the sync). Doing it on every deploy would kill every in-flight turn each time, which is exactly what the detached-scope design exists to avoid, so the one-time cost is paid once at rollout instead.

tella-butler:
<details>

Safe once the two P1 credential-exposure paths below are fixed. The prompter-only authorization and AppArmor/systemd escape fixes otherwise hold across the reviewed launch and resume paths, but OAuth stdio executable resolution and the service installer still leave ways for an existing agent process to capture protected credentials.
_2 inline comments below._
<sub>Reviewed `7a9fb8c` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-99-review)</sub>

</details>

soutar:
Addressing the latest review verdict in `d0ba933d`. Both findings were reproduced on the deployment host before fixing.

**P0, Docker daemon escape.** Confirmed exploitable here, not theoretical: the service user is in the `docker` group and the daemon answers (`ServerVersion 29.1.3`), so a confined agent could ask that root daemon to bind-mount `/var/lib/opensession` or the service user's home into a container and read both the key and the grant store. The caller's path denies do not constrain filesystem work the daemon performs on its behalf, exactly as reported. The docker, containerd, podman and buildkit control sockets are denied from the profile, and `deny /{usr/,}bin/docker x` is added as a cheap extra rather than as the mechanism. Container work stays coordinator-side (`src/server/sandbox`), which is unconfined by design. Profile still parses under `apparmor_parser -Q`, and the two socket denies are pinned in `agent-runtime-security.test.ts`.

**P1, installer path.** Correct, the pre-mount retirement existed only in `deploy/deploy.sh`. `scripts/lib/service.ts` now reads the *installed* unit before overwriting it and, only on the first start that gains `LoadCredential`, stops surviving `opensession-oc-*` and `opensession-preview-*` scopes before the copy. Same narrowing as the deploy path: doing it on every install would kill every in-flight turn, which is what the detached-scope design exists to prevent.

Standing note on the tradeoff, since it now applies twice: closing these brokers removes capabilities agents on this instance genuinely use. `systemd-run --user` and `docker` from an agent shell are both denied after this, and several documented workflows start dev servers, tunnels and browser rigs that way. That is a deliberate call in favour of a boundary that actually holds, and it is one commit to revert if the operator would rather keep the capability.

packages/core/opensession-server/src/server/mcp-oauth-proxy.ts
```diff
@@ -0,0 +1,189 @@
...
+    // the one token it was legitimately handed.
+    const [command, ...args] = secureAgentCommand([
+      String(cfg.command),
+      ...(Array.isArray(cfg.args) ? cfg.args.map(String) : []),
+    ]);
+    transport = new StdioClientTransport({
+      command,
+      args,
+      env: {
+        ...getDefaultEnvironment(),
```
tella-butler: 🔴 **P1** — Do not resolve OAuth stdio commands through agent-writable PATH entries

The new env binding still does not bind the executable that actually receives the token. Supported configs use bare commands such as `bun`, `bunx`, or `npx`, while `opensession.service` deliberately puts service-user-writable directories such as `/home/ubuntu/.bun/bin` and `/home/ubuntu/.local/bin` first in `PATH`. A model shell runs under the same user and the AppArmor profile permits writes there, so it can replace `~/.bun/bin/bunx` after the grant is created. The configured `command`, `args`, and `cfg.env` remain unchanged, `mcpOauthBindingMatches` succeeds, and this spawn hands `SLACK_BOT_TOKEN` to the replacement executable. Require OAuth stdio servers to use an attested, absolute executable and execution chain that the agent cannot modify, or reject stdio OAuth configurations that resolve through service-user-writable binaries/package launchers.

scripts/lib/service.ts
```diff
@@ -215,13 +215,68 @@ export async function install(unitPath?: string): Promise<boolean> {
...
+          "AppArmor is unavailable; personal MCP connections will stay disabled",
+          "set OPENSESSION_PERSONAL_MCP=0 if this host already has an OAuth grant store",
+        );
+      }
       for (const cmd of [
+        ["sudo", "install", "-d", "-m", "0700", "/var/lib/opensession"],
+        [
+          "sudo",
+          "bash",
+          "-c",
```
tella-butler: 🔴 **P1** — Retire unconfined survivors in the service installer too

The new pre-mount retirement in `deploy/deploy.sh` covers only that deployment entry point. `opensession service install` can also upgrade an already-active pre-PR installation: `wasActive` is true, this loop loads the profile and then copies/restarts the unit containing `LoadCredential`, but existing detached `opensession-oc-*` and `opensession-preview-*` scopes survive that restart and remain unconfined. Before boot adoption stops them, such a process can poll `/proc/<new-coordinator-pid>/root/run/credentials` and copy the encryption key and plaintext store. Apply the same first-credential-bearing scope retirement here before copying or starting the new unit.

tella-butler (commented):
OS review · `7a9fb8c`

tella-butler:
<details>

Not safe to merge: the encryption, proxy, and prompter-authorization design is sound, but three P1 escape paths remain. A personal stdio token can still reach a PATH-hijacked executable, and first-install retirement can silently fail or leave detached Pi run hosts alive when the credential mount appears.
_3 inline comments below._
<sub>Reviewed `d0ba933` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-99-review)</sub>

</details>

deploy/deploy.sh
```diff
@@ -126,6 +161,26 @@ while :; do
...
+# (retirePreConfinementRecords) still runs as defence in depth, but it can only
+# run after the mount already exists. Skipped on every later deploy, so
+# detached engines keep surviving restarts the way they are designed to.
+if [ "$CREDENTIAL_ALREADY_MOUNTED" = no ] && [ -s /var/lib/opensession/mcp-oauth.key ]; then
+  echo "[deploy] first credential-bearing start — retiring pre-confinement agent scopes"
+  user_systemctl() {
+    run_as_service_user env XDG_RUNTIME_DIR="/run/user/$SERVICE_UID" systemctl --user "$@"
+  }
+  stale_scopes="$(user_systemctl list-units --plain --no-legend --type=scope \
+    'opensession-oc-*.scope' 'opensession-preview-*.scope' 2>/dev/null | awk '{print $1}' || true)"
```
tella-butler: 🔴 **P1** — Retire detached Pi run hosts before mounting the key

The first-start sweep only lists `opensession-oc-*` and `opensession-preview-*` user scopes. `launchHostUnit()` creates detached local Pi hosts as system units named `bks-run-${hostId}` (`src/server/host-client.ts:337-368`) specifically so they survive coordinator restarts. On a forced deploy after the eight-minute drain cap, a host launched from pre-PR code can remain mid-turn, and its old `makePiBashTool` still spawns unconfined shells. Once the restarted coordinator has `LoadCredential`, that shell runs as the same uid and can read `/proc/<coordinator>/root/run/credentials/.../mcp-oauth-key`, defeating the encrypted store. Stop all live `bks-run-*` system units in both deployment paths before the first credential-bearing start.

scripts/lib/service.ts
```diff
@@ -215,13 +215,68 @@ export async function install(unitPath?: string): Promise<boolean> {
...
+          : []),
+        ...(credentialAlreadyMounted
+          ? []
+          : [
+              [
+                "bash",
+                "-c",
+                "systemctl --user list-units --plain --no-legend --type=scope " +
+                  "'opensession-oc-*.scope' 'opensession-preview-*.scope' 2>/dev/null " +
+                  "| awk '{print $1}' | xargs -r -n1 systemctl --user stop || true",
```
tella-butler: 🔴 **P1** — Fail closed when installer retirement cannot reach the user manager

This retirement command relies on the caller's ambient user bus and suppresses every failure with `2>/dev/null` and `|| true`. The same module documents that `systemctl` can fail with `Failed to connect to bus`; running the installer through root/sudo or an SSH environment without the service user's `XDG_RUNTIME_DIR` therefore yields no units and still proceeds to install and start the credential-bearing unit. Existing unconfined scopes can then read the key through the coordinator's `/proc/.../root` path. Resolve the actual service account, use `/run/user/<uid>` explicitly, and abort before installing the unit unless listing and every required stop succeed.

src/server/mcp-oauth-proxy.ts
```diff
@@ -0,0 +1,189 @@
...
+    if (!preset?.envVar || !cfg.command) {
+      throw new Error("Personal stdio connection has no protected OAuth preset");
+    }
+    // The upstream server is a third-party package that receives the decrypted
+    // token, and it is spawned BY THE COORDINATOR, which is unconfined and
+    // holds the credential mount. Without this wrapper a compromised MCP
+    // package could read the key and every teammate's grant rather than only
+    // the one token it was legitimately handed.
+    const [command, ...args] = secureAgentCommand([
+      String(cfg.command),
```
tella-butler: 🔴 **P1** — Pin the executable, not only the stdio command string

`mcpOauthBindingMatches()` now pins the configured environment, but this still launches the literal `cfg.command` through `getDefaultEnvironment()`'s PATH. The checked-in Slack catalog fixture uses `command: "bunx"` (`src/server/effective-config.test.ts:14`), while the installed service PATH starts with service-user-writable home directories. A model run can replace `~/.bun/bin/bunx` without changing `mcp-config.json` or its pinned environment; the next personal call resolves that replacement and hands it `SLACK_BOT_TOKEN`. `secureAgentCommand()` limits subsequent access but cannot retract the token already delivered. Resolve and bind the launch target to immutable coordinator-managed code, then invoke that exact target rather than searching an agent-writable PATH.

tella-butler (commented):
OS review · `d0ba933`

tella-butler:
<details>

Not safe to merge. The proxy and prompter-scoping changes keep personal tokens out of engine configuration, but current HEAD removes the runtime boundary and leaves the decryption key readable by the same ordinary agent shells this PR is meant to defend against. OAuth stdio launch also remains vulnerable to command capture through the inherited, agent-writable PATH.
_2 inline comments below._
<sub>Reviewed `0c28b99` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-99-review)</sub>

</details>

packages/core/opensession-server/src/server/mcp-oauth.ts
```diff
@@ -8,26 +8,157 @@
...
+        "The mcp-oauth-key credential has an invalid length; expected 32 bytes.",
+      );
+    } catch (error) {
+      // A credential directory without OUR credential in it is the ordinary
+      // case on a host that mounts some other credential, so fall through to
+      // the key file rather than failing the whole feature.
+      if ((error as NodeJS.ErrnoException)?.code !== "ENOENT") throw error;
+    }
+  }
+  const path = statePath(KEY_NAME);
```
tella-butler, line 76: 🔴 **P1** — Keep the encryption key outside the agent's Unix identity

`statePath(KEY_NAME)` places the fallback key under the same state root as the encrypted store, normally `~/.opensession-mcp-oauth.key` beside `~/.opensession-mcp-oauth.json`. The final installer supplies no isolated credential, and model-controlled OpenCode/Pi shells run as the coordinator's Unix user, so a prompt can read both files and decrypt every access and refresh token offline. For example, it can recover Michiel's Slack user token and post as him; mode `0600` does not separate same-UID processes, and a home backup or snapshot also contains both key and ciphertext. This is the original threat the PR claims to close. Do not enable personal MCP with a same-user key fallback: require a broker/second UID or another enforced boundary that keeps both key and store unreadable to model-controlled processes, and fail closed until that boundary exists.

src/server/mcp-oauth-proxy.ts
tella-butler: 🔴 **P1** — Do not resolve OAuth stdio commands through agent-writable PATH entries

`mcpOauthBindingMatches` pins only the literal configured command, arguments, and explicit `cfg.env`, while this launch adds `getDefaultEnvironment()` and lets `StdioClientTransport` resolve the bare command. The codebase represents Slack as `command: "bunx"`, and the deployed PATH resolves `bun`/`bunx` through same-user-writable locations such as the checkout's `node_modules/.bin` and `~/.bun/bin`. An agent can shadow or replace that executable without changing `mcp-config.json`; the binding still matches and this proxy injects the person's Slack token into the attacker's process. Resolve and pin an approved absolute executable from an agent-immutable installation, or reject personal OAuth for stdio configurations that cannot provide that guarantee.

tella-butler (commented):
OS review · `0c28b99`

src/server/mcp-oauth-proxy.ts
tella-butler: Fixed in 2efb367bb, and this one was a live token-capture path independent of the key discussion, so thank you. Confirmed on this host: `/home/ubuntu/.bun/bin` is `drwxrwxr-x ubuntu ubuntu`, so shadowing the name needed no config change at all. The binding now records the command resolved to an absolute path at grant time, the match re-resolves and compares, and the proxy launches the pinned path rather than the name; a binding from before this fails closed and asks for a reconnect. One trap worth recording: `Bun.which()` reads the PATH captured at process start rather than the live one, so the resolution passes PATH explicitly. Without that the pin would ignore the environment the server actually runs with, and the regression test that shadows a name on PATH would have passed for the wrong reason.

packages/core/opensession-server/src/server/mcp-oauth.ts
tella-butler: Accurate, and deliberate. Same-uid readability is stated as a non-goal in the PR body and in `docs/security-model.md` rather than claimed as closed, so this is not a gap between what the code does and what it says.

Declining the recommendation to fail closed, because of what it would cost. Requiring an isolated credential means personal MCP is unavailable on every rootless install, which is precisely the audience of the simple-mode installer in #100: a release artefact under a `systemd --user` service with no root step. Worth noting that an earlier revision of this PR did require `LoadCredential`, and against #100's user-scope unit rendering that is not a degradation but a startup failure, since the unit would carry a credential pointing at a root-owned file its own manager cannot read.

What the fallback key buys is narrower than a boundary and is described as such: the store is ciphertext at rest, so a stray copy, a paste of its contents or a partial sync yields nothing, and it is the substrate a per-use broker sits on. Your backup point was right and the docs overclaimed, so 2efb367bb corrects it: with the fallback key both files live in the same directory, and a whole-home backup can be decrypted offline.

The real fix is the one you name, a broker or a second uid, and it is recorded in `docs/security-model.md` as the intended end state rather than left implicit. It is out of scope here on a considered call by the repository owner: an earlier revision of this PR tried to reach that property with an AppArmor profile, and two review rounds showed it could not hold at a shared uid, while costing agents `systemd-run --user` and `docker`.

packages/core/opensession-server/src/server/mcp-oauth-proxy.ts
tella-butler: Fixed in 2efb367bb (same finding as the later thread on this file). The binding now pins the command RESOLVED to an absolute path at grant time, the match re-resolves and compares it, and the proxy launches that pinned path instead of the configured name, so PATH order no longer decides which binary receives the token. Bindings predating this fail closed and ask for a reconnect. Confirmed your premise on this host: `/home/ubuntu/.bun/bin` is `drwxrwxr-x ubuntu ubuntu`.

src/server/mcp-oauth-proxy.ts
tella-butler: Fixed in 2efb367bb. Agreed the env pin alone did not bind the executable. The stdio binding now records the resolved absolute path, and the proxy invokes exactly that rather than searching PATH; the check re-resolves and refuses when it no longer agrees. Also correct that `secureAgentCommand` was no answer to this, since a token already handed to the process cannot be retracted. That wrapper is gone in any case: the confinement layer was removed at the repository owner's direction because it could not hold at a shared uid, which is described in the updated PR body and in `docs/security-model.md`.

scripts/lib/service.ts
tella-butler: No longer applicable as of the trim (`0c28b99`): `scripts/lib/service.ts` is reverted to main. This PR no longer creates a key, installs an AppArmor profile, or adds `LoadCredential` to the unit, so the installer does not produce a credential-bearing start for surviving scopes to race. The finding was correct against the revision it was written for.

scripts/lib/service.ts
tella-butler: No longer applicable as of the trim (`0c28b99`): the retirement command and the rest of the installer changes are reverted. Recording the substance because it was right and would apply to any future version of this: a retirement that leans on the caller's ambient user bus, and swallows failure with `2>/dev/null` and `|| true`, silently degrades to a no-op under sudo or a bare `ssh host cmd`, which is exactly when it most needs to work.

deploy/deploy.sh
tella-butler: No longer applicable as of the trim (`0c28b99`): `deploy/deploy.sh` is reverted and there is no credential-bearing start to sequence around. Worth recording that the underlying observation was correct and I had missed it: detached Pi run hosts are system units named `bks-run-<hostId>` (`host-client.ts`), so a sweep over `opensession-oc-*` and `opensession-preview-*` user scopes would not have covered them.

tella-butler:
<details>

Not safe to merge: the two P1 credential-exfiltration paths from the previous review remain. The new commit correctly propagates personal MCP scope and pins an executable pathname, but that executable remains writable by the agent identity, while rootless encryption stores its key under the same identity. A run can still recover or capture reusable personal tokens.
_2 inline comments below._
<sub>Reviewed `2efb367` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-99-review)</sub>

</details>

packages/core/opensession-server/src/server/mcp-oauth-proxy.ts
```diff
@@ -0,0 +1,189 @@
...
+    // name would be resolved through the inherited PATH at spawn time, and on
+    // a normal install that runs through directories this same user can write.
+    // mcpOauthBindingMatches has already checked that today's resolution still
+    // agrees with what was pinned when the grant was issued.
+    const command = mcpOauthStdioCommand(name);
+    if (!command) {
+      throw new Error("Personal stdio connection has no pinned executable");
+    }
+    transport = new StdioClientTransport({
+      command,
```
tella-butler: 🔴 **P1** — Do not execute a writable artifact merely because its pathname is pinned

The newest fix pins only the resolved pathname, not the executable's identity or integrity. The codebase's Slack configuration uses `bunx slack-mcp`, and on the deployment host `bunx` resolves to `/home/ubuntu/.bun/bin/bunx`, owned and writable by the same `ubuntu` identity that runs agents. An agent can replace that file in place; `Bun.which("bunx")` still returns the stored pathname, `mcpOauthBindingMatches` still succeeds, and this line launches the replacement with the personal Slack token in its environment. Require an immutable executable and immutable loaded entrypoint, such as a fixed root-owned packaged Slack MCP binary, or move stdio OAuth execution behind the separate-uid credential broker. A pathname comparison alone cannot establish that property.

tella-butler (commented):
OS review · `2efb367`

packages/core/opensession-server/src/server/mcp-oauth-proxy.ts
tella-butler: Correct, and unclosable here, so I am not going to attempt it.

The pathname pin was aimed at a narrower thing and does close it: shadowing a bare name by putting a directory earlier on PATH, which needed no write access to any existing file. In-place replacement of the resolved binary is a different attack and the pin does not stop it.

Nor would a stronger pin. Any integrity check I could add stores its reference value under the same identity: a content hash or a signature lives in the grant store, which is encrypted with a key that same identity can read, so an attacker who can rewrite `~/.bun/bin/bunx` can rewrite the expected hash alongside it. That is your closing sentence generalised, and it is why the answer is the one you name rather than a better comparison.

This is the same property as the key finding in the previous round, pointed at the executable instead of the key: a process running as this user can subvert anything owned by this user. It is documented as a non-goal in the PR body and in `docs/security-model.md` rather than claimed as closed, and the intended end state recorded there is the separate-uid broker, which would fix the key and this together. That is scheduled work, not something this change tries to do.

Flagging one thing for whoever picks that up: if the broker is built, stdio OAuth presets are the harder half, because the token has to reach a third-party process that must run somewhere. The clean version is probably the packaged, coordinator-managed entrypoint you suggest, rather than handing the token to whatever the person's PATH happens to resolve.

tella-butler:
<details>

Not safe to merge until the P1 findings below are fixed. The proxy and prompter-identity changes keep grants out of engine config, but the motivating same-UID token-theft path remains open through the key and OAuth stdio launch chain; the latest commit only removes a stray temp file.
_3 inline comments below._
<sub>Reviewed `e811efe` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-99-review)</sub>

</details>

packages/core/opensession-server/src/server/mcp-oauth.ts
```diff
@@ -8,26 +8,157 @@
...
+    return key;
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException)?.code !== "ENOENT") throw error;
+  }
+  // First use: mint one. `wx` so two racing writers cannot each mint a key and
+  // leave the store encrypted under whichever lost, and 0600 from the open
+  // rather than a later chmod, so it is never briefly world-readable.
+  const key = randomBytes(32);
+  try {
+    const fd = openSync(path, "wx", 0o600);
```
tella-butler, line 94: 🔴 **P1** — Keep the encryption key outside the agent's Unix identity

The fallback key is created as a `0600` file owned by the same Unix user that runs every agent. An agent can read both `~/.opensession-mcp-oauth.key` and the encrypted store, then reproduce the documented AES-256-GCM decryption and recover every access and refresh token. This preserves the exact credential-theft capability that motivated the PR; documenting that limitation does not make the encrypted store a security boundary. Keep decryption behind a broker running as a different UID and give agents only short-lived, run-scoped capabilities, or fail closed when that boundary is unavailable.

packages/core/opensession-server/src/server/mcp-oauth.ts
```diff
@@ -56,12 +187,83 @@ interface ServerAuth {
...
+/** Absolute path for a configured stdio command, or undefined when it cannot
+ *  be resolved. Absolute configured commands are taken as-is; a bare name is
+ *  resolved once, here, so the answer is recorded rather than recomputed from
+ *  whatever PATH the launch happens to inherit. */
+function resolveCommandPath(command: string): string | undefined {
+  if (command.startsWith("/")) return existsSync(command) ? command : undefined;
+  // Pass PATH explicitly: Bun.which() otherwise reads the PATH captured at
+  // process start, which would make the pin ignore the environment the server
+  // is actually running with.
+  return Bun.which(command, { PATH: process.env.PATH ?? "" }) ?? undefined;
```
tella-butler: 🔴 **P1** — Do not resolve OAuth stdio commands through agent-writable PATH entries

`resolveCommandPath` trusts the service's normal `PATH`, while `opensession.service` puts `~/.cargo/bin`, `~/.bun/bin`, and `~/.local/bin` before system directories. The codebase's Slack configuration shape uses bare `bunx`; an agent can place a replacement `bunx` in an earlier writable entry before a teammate connects, causing that malicious path to be recorded as the approved binding and later receive the teammate's token. Resolve OAuth stdio commands only from operator-controlled, non-agent-writable locations, and reject the connection otherwise.

tella-butler (commented):
OS review · `e811efe`

packages/core/opensession-server/src/server/mcp-oauth-proxy.ts
```diff
@@ -0,0 +1,189 @@
...
+      args: Array.isArray(cfg.args) ? cfg.args.map(String) : [],
```
tella-butler: 🔴 **P1** — Do not execute a writable artifact merely because its pathname is pinned

The binding verifies only path and argument strings, not the artifacts they execute. After a legitimate grant pins `/home/ubuntu/.bun/bin/bunx` with `slack-mcp`, an agent running as the same owner can replace that executable or the package/script it loads; the pathname and arguments still match, so the proxy launches attacker-controlled code with `SLACK_BOT_TOKEN` in its environment. Require the full stdio execution chain to be immutable to the agent identity, or move stdio launch and token injection into the separate-UID broker.

tella-butler:
<details>

Safe once the P1 below is fixed. The coordinator proxy now keeps OAuth tokens out of engine configuration and consistently scopes personal grants to the prompter. I retract the prior key-location blocker because the revised description and security model explicitly limit encryption to at-rest protection, but the stdio proxy can still disclose a grant through a mutable pinned executable.
_1 inline comment below._
<sub>Reviewed `a5875d9` · Pi · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-99-review)</sub>

</details>

packages/core/opensession-server/src/server/mcp-oauth.ts
soutar: Fixed in 05c578aca: personal grants are no longer injected into stdio MCP processes at all. Stdio servers keep only their configured workspace credential, so no PATH-resolved executable receives the personal token.

packages/core/opensession-server/src/server/mcp-oauth-proxy.ts
soutar: Fixed in 05c578aca: the coordinator proxy is now HTTP-only and contains no stdio launch path. Personal stdio grants may still power coordinator-owned API actions, but are never handed to the mutable executable or an agent runtime.

packages/core/opensession-server/src/server/mcp-oauth.ts
soutar: This remains deliberately out of scope for the rootless simple-mode target. The fallback key is now described only as protection for a stray copy of the grant file, not as a same-UID boundary or whole-backup protection. Requiring a separate-UID broker would disable personal connections on the default unprivileged install; that broker remains the documented end state rather than a claim made by this PR.

packages/core/opensession-server/src/server/mcp-oauth.ts
```diff
@@ -283,6 +518,7 @@ async function ensureServerAuth(
     ...(scopes ? { scopes } : {}),
     endpoints,
     clientInfo: { clientId: reg.client_id },
+    binding: configuredBinding(name, serverUrl),
```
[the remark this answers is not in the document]
soutar: Fixed in 5136a67e9: re-registering a server at a different URL now starts from a clean registration and does not carry shared or per-user grants across the binding change. Added a regression test proving the old grant is absent and cannot be proxied while the replacement consent flow is pending.

packages/core/opensession-server/src/server/mcp-oauth.ts
```diff
@@ -743,11 +1079,40 @@ export function mcpUserGrantToken(
...
+
+/** Whether this grant can be mounted through the coordinator's HTTP proxy.
+ * Personal grants for stdio tools may still power coordinator-owned UI actions,
+ * but are never handed to the mutable local executable or an agent runtime. */
+export function hasMcpOauthProxyGrantForUsers(
+  name: string,
+  users: Array<string | undefined>,
+): boolean {
+  return (
+    tryReadStore()[name]?.binding?.kind === "http" &&
```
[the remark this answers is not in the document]
soutar: Fixed in 5136a67e9: manual-token saves now require the configured HTTP URL to match, persist that binding, and discard grants from an older URL before installing the newly validated token. Added coverage that the grant is proxy-mounted only for the configured URL and a repointed mismatch fails closed.

packages/core/opensession-server/src/server/interactive-mcp.ts
```diff
@@ -145,6 +178,7 @@ export function interactiveMcpServers(
 			createdBy,
 			isAdmin: true,
 		}),
+		...personalMcp,
```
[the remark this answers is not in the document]
tella-butler: Fixed in 9c7aed202: the Pi MCP runtime now removes external entries shadowed by coordinator-owned in-process servers, and its external connector no longer selects or relays creator-first personal grants. Same-named personal proxies therefore own the runtime entry and use only the prompter identity supplied by interactive-mcp. Added coverage for in-process shadowing.

packages/core/opensession-server/src/server/mcp-oauth.ts
```diff
@@ -99,17 +266,79 @@ function writeGrant(entry: ServerAuth, slot: GrantSlot, grant: Grant): void {
...
+      console.warn(
+        "[mcp-oauth] personal connections are unavailable; the encrypted store was left unchanged",
+      );
+      audit({ kind: "mcp_oauth_store_unavailable" });
+    }
     return {};
   }
 }
 
+export function mcpOauthBindingMatches(
```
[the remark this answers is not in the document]
tella-butler: Fixed in 9c7aed202: both coordinator catalog and tool-call paths now select a personal OAuth header only when the current configured HTTP URL matches the stored grant binding, otherwise falling back to the configured workspace Authorization header. Added matching and repointed-URL coverage.

PR #100, opened 2026-08-18
Simple mode: one-command install of a compiled single-executable

soutar:
## Simple mode: one-command install to a first turn on a fresh box

Open Session's install is an operator's install: a source clone plus `bun install`, four chained `curl | bash` installers, a ten-question wizard, an out-of-band credential step, an opt-in root service, and a minute of silent engine bootstrap before the first turn. Someone who wants to *try it* on a laptop or a cheap VPS has to make a dozen decisions they do not yet have context for.

This adds the simple-mode tier described in `adrs/simple-mode.md` (requirements + definition of done, in the diff) without forking the configurable install: same files, same paths, defaults answered. On a fresh Ubuntu box with only `curl` and `git`:

```
curl -fsSL …/install.sh | bash
```

downloads the published release for the box's OS/arch (no Bun on PATH, no clone), asks nothing, installs and starts a user-level service, and ends with `Open http://127.0.0.1:3850`. Reboot the box and it comes back. First turn (with a model account) is about 13 s on a fresh install, thanks to the prebuilt engine seed.

`--advanced` keeps the full onboarding wizard; `--source` / `--repo` / `--channel` install from a checkout; `opensession service install --system` is the root-unit operator path; a simple install upgrades to a full one with `opensession onboard --force`.

### What changes

**Default install downloads a compiled binary (`install.sh`, `scripts/build-compile.ts`, `.github/workflows/release.yml`)** — the default artefact is a single `bun build --compile` executable, not a Bun + source tree. With no flags the installer resolves `opensession-<os>-<arch>.tar.gz` from the latest GitHub release (`OPENSESSION_RELEASE_BASE` overrides); the tarball holds the target binary, a small `sharp` sidecar `node_modules`, the engine seed, and `release.json`. `--source`/`--repo`/`--channel` install a git checkout + `bun install` (the self-development path). One runner cross-compiles all four targets (`bun build --compile --target=bun-<os>-<arch>`) on a `v*` tag. linux/arm64 artefact is ~72 MB (vs ~203 MB for bun+src). First tag after merge makes the one-command install real end to end.

**Release artefact (`scripts/build-release.ts`)** — `opensession-<ver>-<os>-<arch>.tar.gz`: pinned Bun, prebuilt frontend, `--production` deps pruned of foreign-platform and bundled Claude Code binaries (~500 MB, never executed: every SDK call goes through the installed `claude` CLI), plus `engine-seed/`. 203 MB for linux/arm64. Unpacked under `~/.opensession/releases/<ver>` with a `src` symlink.

**Engine cold start** — opencode's first instance load npm-installs `@opencode-ai/plugin@<its version>` into `~/.config/opencode` (~40 s) and fetches the models.dev catalogue. It skips both when they are already there (`packages/core/src/npm.ts`: `node_modules` exists and the lockfile names the dep; `models-dev.ts` reads the cache and refreshes in the background). So: `package.json opensession.opencodeVersion` pins the engine (passed to opencode's installer), the artefact ships that exact npm tree, `install.sh` copies it in and prefetches `models.json`. The pin is enforced on re-install: `install_cli` reinstalls opencode when the binary's version differs from the pin, and the seed is reconciled to the pinned plugin version rather than only placed when absent. Verified against the opencode binary: a seeded config dir survives an instance load with the lockfile untouched; the harness asserts the same across the first session.

**Installer: no flags, no questions, ends with the URL** — the default path runs `opensession onboard --defaults` (the same wizard with every prompt answered by its fallback: 127.0.0.1:3850, scratch repo, no integrations) and prints the URL last, once `/api/health` answers. It also ensures the service independently of onboarding, which returns early on an existing config — so a re-run recovers a service the first install could not start (for example after the IMDS guard below). `codex` and Tailscale are off the critical path (`--codex`, `--tailscale`); the ChatGPT sign-in reports the install command when `codex` is missing. Onboarding registers a real git scratch repo under `~/.opensession/scratch` when the install is not a checkout, so the first session gets a worktree.

**User-level service (`scripts/lib/service.ts`)** — Linux default is a `systemd --user` unit under `~/.config/systemd/user` with `loginctl enable-linger` (root-free on Ubuntu 24.04; falls back to `sudo -n`, then prints the command). The user-scope rendering drops `User=` (rejected by a user manager), makes the env file optional and targets `default.target`; `renderUnit` uses the release's own `bin/bun` for `ExecStart` so it works regardless of PATH. `status/stop/restart/logs` pick the scope from whichever unit exists and set `XDG_RUNTIME_DIR` themselves so they work from a plain `ssh host cmd`; `loginctl` calls are bounded (a wedged logind otherwise hangs the installer). `--system` renders the `/etc/systemd/system` unit, installs it (the privileged step) before removing any existing user unit, and restores the user unit if a later step fails — so switching scopes on a box without sudo cannot cause an outage, and the two never compete for port 3850.

**Metadata endpoint (IMDS) guard** — the engine detaches each session into its own `systemd --user` scope (so restarts don't kill in-flight turns), which is outside any unit's cgroup — so no unit's `IPAddressDeny=` reaches the agent's children under the user *or* the `--system` unit. On stock Ubuntu a per-user manager cannot apply `IPAddressDeny=` at all (`PrivateUsers=` is denied to unprivileged users, silently). So `service install` probes `169.254.169.254` and refuses when anything answers, printing the controls that actually cover the engine's children: a host firewall rule (`iptables … -m owner --uid-owner <uid> -j REJECT`), `OPENSESSION_OC_DETACH=0` (keep engines in the system cgroup), or `OPENSESSION_ALLOW_IMDS=1` on a box with no role. Documented in `docs/setup/integrations-misc.md`. The pre-existing `sudo systemd-run … curl 169.254.169.254` role mint stays gated on `AGENT_AWS_CREDS`/`AGENT_AWS_REGION`/config.

**`opensession update` for release installs (`scripts/lib/update.ts`)** — a release install (no `.git`, a `src` symlink) updates by download-and-swap instead of a git pull: fetch the latest artefact into a staging dir, rename the unpacked tree into `releases/<name>` only after `release.json` is present (a partial dir from a failed `tar` is discarded and redone, never trusted), reconcile the engine to the new pin BEFORE activating it (reinstall opencode, refresh the plugin seed) and only swap the `src` symlink and restart when every required component matches — a failed reconcile leaves the current release active and returns nonzero, with the downloaded tree kept for a retry. The previous release stays for rollback. `OPENSESSION_RELEASE_BASE` / `--channel <url>` override the source.

**Uninstall (`install.sh --uninstall`)** — removes the service (either scope), the shim, the PATH block from every profile it appended to, skill symlinks into the install, the release/checkout, config, env file and the `~/.opensession-*` state (sessions, accounts, audit); confirmed on a terminal, `--yes` for scripts. State removal classifies by what to keep, not an ever-growing list of what to delete: it removes every `~/.opensession-*` name except the two that can hold user work — a non-empty scratch workspace dir, and the configured worktrees dir (env/config-resolved) when it holds uncommitted or unpushed work (commits on no remote and not on main/master). Those are preserved and named; the install home is kept whole when its own worktrees are unsaved. Repositories registered from elsewhere are never under the install root. `goss.uninstalled` asserts the clean case; a STRICT harness case asserts the preserve case.

### Test harness (`test/simple-mode/`)

A `bun test` driver builds the artefact for the guest's arch, creates a fresh Ubuntu 24.04 VM with Lima, runs `install.sh` in it exactly as a customer would (no flags, no Bun, no clone), waits for the installer's own service, and validates with goss. `SIMPLE_MODE_STRICT=1` adds the definition-of-done checks (user service active, linger, `doctor` clean but for accounts, no tailscale/codex/source checkout), reboots the VM and checks again, then checks the uninstall preserve-case (a committed scratch worktree) and the clean-removal case. `OPENSESSION_TEST_CLAUDE_TOKEN` runs a real turn. `SIMPLE_MODE_SOURCE=1` is the contributor path.

STRICT is 12/12 on a fresh VM installing the compiled binary, including on this merged tree. Not part of the default `bun test` set (minutes, needs Lima); README in the directory.

### Not in this PR (out of scope)
The `/setup` first-run page (ADR R2.2/R2.3 — name → account → repo → first turn) is owned by the first-time-usage work and lives in its own change; this PR stops at "installed and running robustly" and leaves those requirements for it. Also open: R2.4 next-steps checklist in the UI · R3.2 launchd log rotation · R4 webhooks without a public server · R5 maintenance toil · macOS LaunchAgent path is unit-tested only. A follow-up worth calling out: the metadata-endpoint block is a host-firewall concern for detached engine scopes under *either* service scope — this PR surfaces and documents it but does not change the detach model.


tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `816dce9`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)

</details>

tella-butler:
<details>

Not safe to merge. The advertised no-flag install still performs a source build, existing installations are incorrectly routed through first-run setup, and the default user service removes an important IMDS security boundary. Uninstall also leaves several Open Session artifacts behind despite the stated cleanup guarantee.
_4 inline comments below._
<sub>Reviewed `816dce9` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

</details>

tella-butler:
<details>

Not safe to merge. The release and first-run paths substantially simplify a fresh installation, but uninstall can destroy unsaved session work, and reruns do not reliably reconcile or recover the installed runtime. The service scope transition also permits two units to compete for the same server.
_4 inline comments below._
<sub>Reviewed `57173c5` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

</details>

tella-butler (commented):
OS review · `57173c5`

install.sh
```diff
@@ -87,23 +121,101 @@ if [ "$DO_UNINSTALL" = "1" ]; then
...
+  if [ "$NO_PROMPT" != "1" ]; then
+    remove_data=0
+    if [ -r /dev/tty ] && { : </dev/tty; } 2>/dev/null; then
+      printf '  remove %s, %s and %s/.opensession-* (sessions, accounts, audit)? [y/N] ' "$OPENSESSION_HOME" "$HOME/.opensession.env" "$HOME"
+      read -r answer </dev/tty || answer=""
+      case "$answer" in y*|Y*) remove_data=1 ;; esac
+    fi
+  fi
+  # Never delete an agent's unsaved work. Session worktrees live under
+  # $OPENSESSION_HOME/worktrees, so `rm -rf $OPENSESSION_HOME` would take any
```
tella-butler: 🔴 **P0** — Uninstall deletes unsaved session work

Simple-mode worktrees default to `$OPENSESSION_HOME/worktrees`, and this recursively removes the entire home without inspecting those Git repositories. A code session can legitimately leave uncommitted files or commits that have not reached any remote; `install.sh --uninstall --yes` then destroys that work irreversibly. This conflicts with the codebase's existing worktree cleanup invariant, which preserves or banks dirty and unpushed work. Before removing the home, enumerate its worktrees and keep the home if any tree is dirty or has commits unreachable from a remote, naming the paths that blocked removal.

scripts/lib/service.ts
```diff
@@ -211,28 +344,79 @@ export function renderPlist(): string {
...
+          return false;
+        }
+        mkdirSync(dirname(USER_UNIT_PATH), { recursive: true });
+        await Bun.write(USER_UNIT_PATH, unit);
+        info(dim(`installed ${USER_UNIT_PATH}`));
+        await enableLinger();
+        // Enabling linger starts user@<uid>.service; from a session that had
+        // no user manager (plain ssh command, cron), give its bus a moment to
+        // appear before the first `systemctl --user` call.
+        const runtimeDir = process.env.XDG_RUNTIME_DIR ?? env.XDG_RUNTIME_DIR ?? "";
```
tella-butler, line 437: 🟠 **P2** — Switching to system scope leaves the user unit running

The user branch rejects an existing system unit, but the system branch has no symmetric check for `USER_UNIT_PATH`. After a normal simple install, `opensession service install --system` therefore enables a second `opensession.service` while the user unit still owns port 3850; the system unit crash-loops or fails, and `installedScope()` continues selecting the user unit so status, logs, and uninstall hide the system unit. Either migrate atomically by stopping/removing the user unit first, or reject the command with instructions to uninstall the existing scope.

tella-butler:
<details>

Not safe to merge: the revised uninstall guard still permits permanent session-work loss, and release-update retries can activate an incomplete release. The previously reported system/user service collision also remains.
_3 inline comments below._
<sub>Reviewed `6b38397` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

</details>

install.sh
```diff
@@ -87,23 +121,101 @@ if [ "$DO_UNINSTALL" = "1" ]; then
...
+  # it has nowhere to push). If anything is unsaved, keep the whole home and
+  # say where — the rest of the removal (service, shim, PATH, env, state)
+  # still runs. Registered repos live outside the home and are never touched.
+  dirty_worktrees=""
+  if [ "$remove_data" = "1" ] && command -v git >/dev/null 2>&1 && [ -d "$OPENSESSION_HOME/worktrees" ]; then
+    while IFS= read -r gitpath; do
+      wt="$(dirname "$gitpath")"
+      [ -n "$(git -C "$wt" status --porcelain 2>/dev/null)" ] && { dirty_worktrees="$dirty_worktrees  $wt (uncommitted)"$'
+'; continue; }
+      if [ -n "$(git -C "$wt" remote 2>/dev/null)" ] && [ -n "$(git -C "$wt" log --oneline --branches --not --remotes 2>/dev/null)" ]; then
```
tella-butler: 🔴 **P0** — Uninstall still deletes committed work from the scratch repository

The new guard only checks for unpushed commits when `git remote` returns something. The default release install creates `~/.opensession/scratch` with no remote, then code sessions create worktrees from it. If an agent commits changes in one of those worktrees, its status is clean and this condition skips the commit check because the repository has no remote. `install.sh --uninstall --yes` then removes `~/.opensession`, destroying the only copy of those commits. Treat no-remote worktrees as unsafe when they contain commits beyond their base branch, or conservatively preserve the install home whenever a no-remote session worktree exists. Add a harness case with a clean committed scratch branch.

scripts/lib/update.ts
```diff
@@ -26,14 +26,114 @@
...
+
+    const releasesDir = join(OPENSESSION_HOME, "releases");
+    mkdirSync(releasesDir, { recursive: true });
+    const dest = join(releasesDir, relName);
+    const current = existsSync(rel.srcLink) ? realpathSync(rel.srcLink) : "";
+    if (existsSync(dest) && realpathSync(dest) === current && current) {
+      ok("already up to date", rel.manifest.version);
+      return 0;
+    }
+    if (!existsSync(dest)) {
```
tella-butler: 🔴 **P1** — A failed extraction is trusted as a complete release on retry

Extraction writes directly to the final `releases/<relName>` directory. If `tar` fails partway through, for example because the disk fills, it returns an error but leaves that partial directory behind. After the operator frees space and reruns `opensession update`, `existsSync(dest)` is true, so the updater skips extraction, switches `src` to the incomplete tree, and restarts into missing or truncated files. Extract into a temporary sibling directory, validate at least `release.json` and the required entrypoints, then atomically rename it to `dest`; remove the staging directory on every failure.

scripts/lib/service.ts
```diff
@@ -211,28 +344,79 @@ export function renderPlist(): string {
...
+        for (let i = 0; i < 20 && !existsSync(join(runtimeDir, "systemd", "private")); i++) {
+          await Bun.sleep(500);
+        }
+      } else {
+        const staged = join(OPENSESSION_HOME, "opensession.service");
```
tella-butler, line 448: 🟠 **P2** — Switching to system scope still leaves the user service installed

When `opensession service install --system` runs after the default user service, `installedScope()` returns `user`, but the system branch proceeds without stopping or removing that unit. The new system unit then competes for the same port, normally fails to start, and all later status/control operations continue targeting the user unit because user scope wins detection. Refuse this transition with an actionable error, or disable and remove the user unit before enabling the system unit.

tella-butler (commented):
OS review · `6b38397`

tella-butler:
<details>

Not safe to merge. The latest commit fixes committed scratch detection and atomic extraction, but uninstall can still permanently delete configured worktrees, and an update can activate a release with an incompatible engine. System-scope migration also removes the working user service before privileged installation succeeds.
_3 inline comments below._
<sub>Reviewed `e23b83a` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

</details>

install.sh
```diff
@@ -87,23 +121,108 @@ if [ "$DO_UNINSTALL" = "1" ]; then
...
+  fi
+  if [ "$remove_data" = "1" ] && [ -n "$dirty_worktrees" ]; then
+    remove_data=0
+    warn "keeping $OPENSESSION_HOME: session worktrees have unsaved work"
+    printf '%s' "$dirty_worktrees" | sed '/^$/d'
+    muted "commit or push them, or delete $OPENSESSION_HOME by hand once you are sure"
+  fi
+  if [ "$remove_data" = "1" ]; then
+    rm -rf "$OPENSESSION_HOME" "$HOME/.opensession.env"
+    for f in "$HOME"/.opensession-*; do [ -e "$f" ] && rm -rf "$f"; done
```
tella-butler: 🔴 **P0** — Uninstall still deletes configured worktrees without checking them

`onboard --advanced` accepts any `paths.worktreesDir`, and `configuredPaths()` uses that value, but this guard scans only `$OPENSESSION_HOME/worktrees`. For example, an operator can configure `~/.opensession-worktrees`; `--uninstall --yes` will not inspect its dirty worktrees, then this `$HOME/.opensession-*` wildcard permanently deletes them. Read and guard the configured or environment-provided worktree root before deletion, and replace the wildcard with an explicit list of Open Session-owned state paths.

scripts/lib/update.ts
```diff
@@ -26,14 +26,212 @@
...
+    // so a reader either sees the old tree or the new one, never neither.
+    const staging = join(dirname(rel.srcLink), `.src.next.${process.pid}`);
+    try {
+      rmSync(staging, { force: true });
+    } catch {}
+    symlinkSync(dest, staging);
+    renameSync(staging, rel.srcLink);
+    ok(`switched src -> releases/${relName}`, current ? `was ${current.split("/").pop()}` : undefined);
+    try {
+      await reconcileEngine(dest);
```
tella-butler: 🔴 **P1** — Do not activate a release when engine reconciliation fails

The code explicitly treats the pinned engine, plugin, and bundled SDK as one compatibility set that breaks turns when mismatched, but the symlink is swapped before reconciliation and every reconciliation failure is non-fatal. If the opencode installer fails during a temporary network outage, update still restarts the new release and returns success with the old engine. A retry then exits through the `already up to date` branch before reconciliation, so it cannot repair the mismatch. Reconcile successfully before swapping `src`, or roll back and return non-zero; the already-current path must also retry reconciliation.

scripts/lib/service.ts
```diff
@@ -211,28 +350,88 @@ export function renderPlist(): string {
...
+        if (existsSync(USER_UNIT_PATH)) {
+          info(dim("removing the existing user service before installing the system unit"));
+          await runInherit(systemctl("user", ["disable", "--now", SERVICE_NAME]), undefined, env);
+          await run(["rm", "-f", USER_UNIT_PATH]);
+          await runInherit(systemctl("user", ["daemon-reload"]), undefined, env);
+        }
+        const staged = join(OPENSESSION_HOME, "opensession.service");
+        await Bun.write(staged, unit);
+        info(dim(`installing ${staged} -> ${SERVICE_PATH} (needs sudo)`));
+        if ((await runInherit(["sudo", "cp", staged, SERVICE_PATH])) !== 0) {
```
tella-butler, line 451: 🟠 **P2** — System migration removes the working service before sudo succeeds

The migration disables and deletes the existing user unit before attempting this privileged copy. If the operator cancels the sudo prompt, enters a wrong password, or `sudo cp` otherwise fails, the command returns false after taking down the previously working service and leaves no installed unit. Copy the system unit successfully first, then stop and remove the user unit immediately before enabling the system unit; restore the user unit if the later activation fails.

tella-butler (commented):
OS review · `e23b83a`

soutar:
Rewrote the branch into clean diff-state commits (force-push); the tree is unchanged from what was verified. The description reflects the final state. Where the earlier inline findings are addressed:

- **Release download by default / release workflow** — `Install from a published release, no questions, clean uninstall` (+ `.github/workflows/release.yml`).
- **Engine version pin on upgrade** — same installer commit (`install_cli` reinstalls on mismatch, seed reconciled to the pin) and `opensession update for release installs` (reconciles binary + seed after the swap).
- **Legacy/advanced installs never routed to /setup** — `First-run page at /setup`: `done` is true unless `setup.firstRunDone` is explicitly `false`, which only the defaults path writes.
- **Uninstall preserves unsaved work** — installer commit: keeps the home when a worktree has uncommitted changes or commits on no remote and not on main/master; harness asserts both cases.
- **Atomic release extraction** — update commit: staging dir, `release.json` check, rename into place; a partial dir is discarded.
- **System scope migrates off the user unit; IMDS boundary** — `Run as a user-level systemd service, with a metadata-endpoint guard`: `--system` removes the user unit first, `renderUnit` uses the release's own `bin/bun`, and install refuses on a reachable metadata endpoint, naming the host-firewall control that actually covers detached engine scopes.
- **Rerun recovers a failed service install** — installer ensures the service independently of onboarding.

STRICT harness 12/12 on a fresh VM. The three red CI checks are the pre-existing `main` failures (the transcript-snapshot test and `doctor` exiting non-zero under `--no-engine`), not from this branch.

tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `8be5ba1`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)

</details>

tella-butler:
<details>

Not safe to merge because uninstall still has two verified data-loss paths, and release updates can activate an engine-incompatible release while reporting success. The previous partial-extraction and committed-scratch fixes hold, but the remaining destructive and partial-failure behavior must be fixed first.
_4 inline comments below._
<sub>Reviewed `8be5ba1` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

</details>

install.sh
tella-butler: 🔴 **P0** — Uninstall deletes non-Git scratch workspace contents

The safety check only examines Git worktrees, but scratch-mode workspaces use `stateDir("scratch")`, which resolves to `~/.opensession-scratch/<workspace>`, as their writable working directory. These directories are deliberately not Git repositories and persist until the workspace is deleted. For example, a feed workspace containing a downloaded and edited `recording.mp4` passes this guard and is then recursively removed by the `~/.opensession-*` wildcard. Preserve non-empty scratch workspaces or require a separate explicit destructive confirmation that identifies them; do not include them in an unchecked wildcard removal.

install.sh
```diff
@@ -87,23 +121,108 @@ if [ "$DO_UNINSTALL" = "1" ]; then
...
+      # are on no remote and not on the repo's main/master. Deleting the home
+      # deletes the branch, so these would be the only copy. A no-remote repo
+      # (scratch) is judged the same way — commits an agent added on top of the
+      # base branch count, a freshly branched worktree at the base does not.
+      base="$(git -C "$wt" rev-parse --verify -q main 2>/dev/null || git -C "$wt" rev-parse --verify -q master 2>/dev/null || true)"
+      if [ -n "$(git -C "$wt" log --oneline HEAD --not --remotes ${base:+"$base"} 2>/dev/null)" ]; then
+        dirty_worktrees="$dirty_worktrees  $wt (unpushed commits)"$'\n'
+      fi
+    done <<EOT
+$(find "$OPENSESSION_HOME/worktrees" -maxdepth 4 -name .git 2>/dev/null)
```
tella-butler: 🔴 **P0** — The guard misses configured worktree directories

The advanced wizard and `OPENSESSION_WORKTREES_DIR` both allow worktrees outside `$OPENSESSION_HOME/worktrees`, but this scan is hard-coded to that default. A user who configures `/home/alice/.opensession-worktrees` can have uncommitted or unpushed session work there; the scan sees nothing, then the later `$HOME/.opensession-*` wildcard deletes the configured directory. Resolve the effective worktree directory from the environment/config and scan it before removal, and never wildcard-delete an unclassified directory.

scripts/lib/update.ts
tella-butler: 🔴 **P1** — Do not activate a release when engine reconciliation fails

The `src` symlink is switched before reconciliation, and every reconciliation failure is reduced to a warning followed by a successful update and restart. During a release bump from OpenCode 1.18.18 to 1.19.0, for example, a network failure in the installer leaves the old binary in place while the new release and possibly the new plugin runtime are activated. This file explicitly documents that the engine, plugin, and SDK form one compatibility set, so subsequent turns can fail even though `opensession update` returned success. Make reconciliation return a success result and only activate/restart the new release after every required engine component matches, or restore the old symlink and return nonzero.

scripts/lib/service.ts
tella-butler, line 451: 🟠 **P2** — System migration removes the working service before sudo succeeds

The migration stops the user service and deletes its unit before the first privileged operation. If the account is not in sudoers or `sudo cp` otherwise fails, this branch returns false after removing the only working service, so the operator's attempt to switch scopes causes an outage and loses reboot persistence. Copy and validate the system unit first, then perform the stop/start handoff, with rollback that restores and re-enables the user unit if any later systemd step fails.

tella-butler (commented):
OS review · `8be5ba1`

soutar:
All four findings fixed and verified on a VM; the branch stays 11 clean commits (force-push).

- **P0 · uninstall deletes non-Git scratch workspace contents.** Removal now classifies by what to *keep*, not a list of what to delete: it clears every `~/.opensession-*` name except a non-empty `~/.opensession-scratch` (workspace files) and the resolved worktrees dir when it holds unsaved work. Verified: a `~/.opensession-scratch/ws1/recording.mp4` survives; an empty scratch dir and all app state are removed (clean uninstall leaves zero `~/.opensession-*`).
- **P0 · guard misses configured worktree directories.** The scan resolves the effective worktrees dir from `OPENSESSION_WORKTREES_DIR` / config `paths.worktreesDir` (default `~/.opensession/worktrees`) and scans *that*; a dir configured at e.g. `~/.opensession-worktrees` is preserved when it has uncommitted/unpushed work rather than wildcard-deleted. Verified: a custom worktrees dir with a committed no-remote branch survives uninstall.
- **P1 · don't activate a release when engine reconciliation fails.** Reconciliation now runs *before* the symlink swap and returns a result; if opencode can't be brought to the release's pin (or the seed can't be refreshed), update leaves the current release active, keeps the downloaded tree for a retry, and returns nonzero. Verified against a release pinning an unreinstallable opencode: `keeping the current release`, rc=1, `src` unchanged, old server still healthy. (Also hardened the tarball's top-dir detection against an AppleDouble `._` first entry.)
- **P2 · system migration removes the working service before sudo succeeds.** `--system` now installs the system unit (the privileged `sudo cp`) *first* — so a denied sudo returns with the user unit untouched — then stops/removes the user unit, and restores it if a later systemd step fails. Verified: user-active → `--system` → user unit gone, system unit healthy.

Also fixed a regression from merging main: the release build pruned the `sharp` native, but main's new session-social-card feature loads `sharp` at boot; the platform's sharp native is now kept (other platforms' variants still pruned). And two dangling references at main's tip (`findSessionAsync`, `getCachedSessions`) are reconciled so the branch compiles.

STRICT harness 12/12 on a fresh VM (includes the committed-scratch uninstall case). The red CI checks remain the pre-existing `main` failures.

tella-butler:
<details>

Unsafe to merge until the remaining uninstall data-loss path and partial-release retry are fixed. The new commits correctly keep the old release active when engine reconciliation fails and safely restore the user service during system-scope migration, but two release-install failure paths remain.
_2 inline comments below._
<sub>Reviewed `a9ff3ab` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

</details>

install.sh
```diff
@@ -87,23 +121,151 @@ if [ "$DO_UNINSTALL" = "1" ]; then
...
+  scratch_has_data=0
+  [ -d "$scratch_state" ] && [ -n "$(ls -A "$scratch_state" 2>/dev/null)" ] && scratch_has_data=1
+
+  if [ "$remove_data" = "1" ]; then
+    kept_any=0
+    # The install home holds the default worktrees dir and the scratch git repo
+    # its worktrees branch from; keep the whole thing if any of its worktrees is
+    # unsaved. A worktrees dir configured OUTSIDE the home is preserved on its
+    # own in the state sweep below.
+    if [ -n "$dirty_worktrees" ] && case "$worktrees_dir" in "$OPENSESSION_HOME"/*) true ;; *) false ;; esac; then
```
tella-butler, line 233: 🔴 **P0** — External scratch worktrees still lose their Git object store

The new guard still falls short when a release install's configured `worktreesDir` is outside `$OPENSESSION_HOME`. For example, advanced onboarding can keep the scratch repository at `~/.opensession/scratch` while placing its worktrees in `/srv/opensession-worktrees`. If `/srv/opensession-worktrees/scratch-feature` has a unique commit, this scan correctly populates `dirty_worktrees`, but this condition is false because `/srv/opensession-worktrees` is not under `~/.opensession`; line 237 therefore deletes the scratch repository's common Git directory and object database. The external worktree remains, but its `.git` file points into the deleted repository and the unique commit objects are gone. Preserve `$OPENSESSION_HOME` whenever a guarded worktree's Git common directory is inside it, or conservatively whenever `dirty_worktrees` is non-empty.

install.sh
```diff
@@ -227,7 +389,64 @@ install_bun_via_python() {
...
+    http://*|https://*)
+      art_tmp="$(mktemp -d)"
+      curl -fsSL "$ARTIFACT" -o "$art_tmp/release.tar.gz" || die "could not download $ARTIFACT"
+      art_file="$art_tmp/release.tar.gz" ;;
+    *) art_file="$ARTIFACT"; [ -f "$art_file" ] || die "no such file: $art_file" ;;
+  esac
+  # awk reads the whole listing so tar never sees a closed pipe (pipefail).
+  rel_name="$(tar -tzf "$art_file" 2>/dev/null | awk -F/ 'NR==1{print $1}')"
+  [ -n "$rel_name" ] || die "could not read the release tarball"
+  if [ ! -d "$RELEASES/$rel_name" ]; then
```
tella-butler, line 433: 🔴 **P1** — Retrying an interrupted install trusts a partial release

This has the same retry hazard that was fixed in `scripts/lib/update.ts`: extraction writes directly into the final release directory, while subsequent runs treat that directory's existence as proof that extraction completed. If `tar` is killed or the disk fills after creating `~/.opensession/releases/<name>`, the first run exits but leaves that directory. After the transient problem is fixed, every retry prints `already unpacked`, links `src` to the incomplete tree, and can either abort later on a missing `package.json` or start a broken release. Extract into a sibling staging directory, validate the expected manifest/runtime files, and atomically rename it into place only after successful extraction.

tella-butler (commented):
OS review · `a9ff3ab`

tella-butler:
<details>

Not safe to merge. The user-to-system service migration fix holds, but the two previously reported release-install blockers remain: uninstall can destroy the repository backing external scratch worktrees, and installer retries trust incomplete extracts. Engine reconciliation can also damage the active runtime on failure, while the new macOS launcher breaks supported install paths containing spaces.
_4 inline comments below._
<sub>Reviewed `9067d60` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

</details>

tella-butler (commented):
OS review · `9067d60`

scripts/lib/update.ts
```diff
@@ -26,14 +26,238 @@
...
+    "opencode",
+  );
+  if (existsSync(join(seedSrc, "node_modules"))) {
+    const want = seededPluginVersion(seedSrc);
+    const has = seededPluginVersion(ocConfig);
+    if (want && has !== want) {
+      try {
+        mkdirSync(ocConfig, { recursive: true });
+        for (const f of ["node_modules", "package.json", "package-lock.json"]) {
+          rmSync(join(ocConfig, f), { recursive: true, force: true });
```
tella-butler, line 141: 🔴 **P1** — Engine reconciliation failure can corrupt the active runtime

The update mutates shared engine state before it knows reconciliation will finish. For example, when moving from engine A to B, the opencode installer can successfully replace the binary with B, then this block deletes A's plugin runtime and `cpSync` can fail because the disk fills. The function returns false and leaves the old release symlink active, but that release now runs against binary B and a missing or partial plugin tree, which is exactly the incompatible combination this reconciliation is intended to prevent. Back up or atomically stage every engine-owned path, including the existing opencode binary, and restore all of them on any failure before claiming the current release was kept.

scripts/lib/service.ts
```diff
@@ -156,44 +211,143 @@ async function resolveUsername(): Promise<string> {
...
   const bun = bunPath();
-  // launchd has no EnvironmentFile, so source the env file in a shell first.
-  // `set -a` exports everything it defines.
-  const script = `set -a; [ -f ${ENV_PATH} ] && . ${ENV_PATH}; set +a; exec ${bun} run opensession.ts`;
+  return (
+    `#!/bin/bash\n` +
+    `# macOS shows this file's name in Login Items & Extensions; hence "OpenSession".\n` +
+    `cd ${REPO_ROOT} || exit 1\n` +
+    `set -a; [ -f ${ENV_PATH} ] && . ${ENV_PATH}; set +a\n` +
+    `exec ${bun} run opensession.ts\n`
```
tella-butler: 🟠 **P2** — The macOS launcher does not quote filesystem paths

The generated shell script interpolates `REPO_ROOT`, `ENV_PATH`, and the Bun path as unquoted shell words. A supported source install such as `install.sh --dir "/Users/alice/Open Session"` therefore generates `cd /Users/alice/Open Session` and an equally broken `exec` command. The LaunchAgent bootstraps successfully but its launcher immediately exits and KeepAlive repeatedly restarts it. Shell-quote each interpolated path, including embedded quotes and shell metacharacters, before writing the launcher.

soutar:
Scope: dropped the `/setup` first-run page (name → account → repo → first turn) — that's owned by the first-time-usage work and belongs in its own change. This PR now stops at "installed and running robustly." Removed: `FirstRun.tsx`, the `/setup` route/redirect, the `/api/setup/first-run` endpoints, the `firstRunDone` config marker, and the `/setup` goss checks. Kept as they stand on their own: the engine endpoint reporting its real enabled state, the Workspace → Usage account pointers, and the session-cache merge reconcile. History rebuilt into 11 clean commits (force-push); no first-run leftovers; typecheck + 130 unit tests green; `mergeable`.

tella-butler:
<details>

Not safe to merge. Although the sudo-ordering migration issue is fixed, the latest tree drops the promised browser first-run flow and still has uninstall/install/update paths that can lose Git history or leave the runtime unusable; the macOS service also fails for supported install paths containing spaces.
_5 inline comments below._
<sub>Reviewed `1289f50` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

</details>

tella-butler (commented):
OS review · `1289f50`

adrs/simple-mode.md
```diff
@@ -0,0 +1,352 @@
...
+
+## R2. First run in the browser, not the terminal
+
+- **R2.1** The installer writes `config.json` + `.env` with all defaults
+  (product name, `127.0.0.1:3850`, webhook `3848`, worktrees dir, no
+  integrations, no automations) and starts the server. `opensession onboard`
+  becomes `--advanced` mode; the ten questions still exist there. The
+  no-flag installer runs `onboard --defaults` (no questions, service
+  installed, ends with the URL).
+- **R2.2** `/setup` first-run page, served until complete:
```
tella-butler, line 104: 🔴 **P1** — The default install no longer reaches the promised first turn

The current tree has no `/setup` route, pending marker, account/repository sequence, or browser-triggered first turn. `onboard --defaults` writes an empty team, so opening the printed URL reaches the ordinary `UserGate`, which offers the hardcoded `Local User`, then drops into the normal app. The new harness likewise removed its `/setup` assertions and proves a turn only by creating one directly with `curl`. Consequently the advertised fresh-box path does not satisfy R2.2/R2.3 or the PR description. Restore the first-run gate/page/API and its end-to-end assertion, or explicitly narrow the PR and its stated definition of done to distribution and service installation.

scripts/lib/service.ts
```diff
@@ -156,44 +211,143 @@ async function resolveUsername(): Promise<string> {
...
+ */
+export function renderLauncher(): string {
   const bun = bunPath();
-  // launchd has no EnvironmentFile, so source the env file in a shell first.
-  // `set -a` exports everything it defines.
-  const script = `set -a; [ -f ${ENV_PATH} ] && . ${ENV_PATH}; set +a; exec ${bun} run opensession.ts`;
+  return (
+    `#!/bin/bash\n` +
+    `# macOS shows this file's name in Login Items & Extensions; hence "OpenSession".\n` +
+    `cd ${REPO_ROOT} || exit 1\n` +
```
tella-butler, line 357: 🟠 **P2** — The macOS launcher does not quote filesystem paths

All three interpolated paths in the generated shell script are unquoted. A supported source install such as `--dir '/Users/alice/Open Session'`, or an overridden home under `Library/Application Support`, generates commands like `cd /Users/alice/Open Session` and `exec /Users/alice/Open Session/bin/bun ...`; Bash splits these into separate arguments and the LaunchAgent enters a restart loop. Shell-quote `REPO_ROOT`, `ENV_PATH`, and the Bun path when rendering the launcher.

tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `83c78a8`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)

</details>

tella-butler:
<details>

Not safe to merge. The service migration fixes and delayed `src` activation address several prior findings, but uninstall can still destroy Git history, and release installation and updates still have partial-failure paths that leave unusable runtimes. The implementation also no longer delivers the first-run and first-turn flow that defines this PR's stated goal.
_4 inline comments below._
<sub>Reviewed `4fa4d69` · Pi · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

</details>

tella-butler:
<details>

Not safe to merge until the data-loss path and release update failures below are fixed. The latest commits correctly repair the service-scope migration ordering, but uninstall can still destroy the backing repository for preserved worktrees, interrupted installs remain unrecoverable, and engine updates are not transactional. The documented first-turn scope also still exceeds what this PR now delivers.
_4 inline comments below._
<sub>Reviewed `4fa4d69` · Pi · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

</details>

install.sh
```diff
@@ -87,23 +122,151 @@ if [ "$DO_UNINSTALL" = "1" ]; then
...
+  scratch_has_data=0
+  [ -d "$scratch_state" ] && [ -n "$(ls -A "$scratch_state" 2>/dev/null)" ] && scratch_has_data=1
+
+  if [ "$remove_data" = "1" ]; then
+    kept_any=0
+    # The install home holds the default worktrees dir and the scratch git repo
+    # its worktrees branch from; keep the whole thing if any of its worktrees is
+    # unsaved. A worktrees dir configured OUTSIDE the home is preserved on its
+    # own in the state sweep below.
+    if [ -n "$dirty_worktrees" ] && case "$worktrees_dir" in "$OPENSESSION_HOME"/*) true ;; *) false ;; esac; then
```
tella-butler, line 233: 🔴 **P0** — Preserved external worktrees still lose their backing Git repository

This keeps `$OPENSESSION_HOME` only when `worktrees_dir` is inside it. In a default release install, the configured scratch repository is `~/.opensession/scratch`; if `paths.worktreesDir` is `/srv/opensession-worktrees`, each external worktree's `.git` file points back into that repository. With an unpushed commit, this scan marks the external worktree dirty, but the condition is false and `rm -rf "$OPENSESSION_HOME"` deletes its object store. The worktree directory remains, but Git can no longer resolve its commit or index, causing the exact data loss the guard is intended to prevent. Keep the home whenever any protected worktree is backed by a repository under it. The smallest safe fix for the current layout is to keep the home whenever `dirty_worktrees` is non-empty, regardless of where the worktree directories live.

install.sh
```diff
@@ -227,7 +390,69 @@ install_bun_via_python() {
...
+    *) art_file="$ARTIFACT"; [ -f "$art_file" ] || die "no such file: $art_file" ;;
+  esac
+  # awk reads the whole listing so tar never sees a closed pipe (pipefail).
+  rel_name="$(tar -tzf "$art_file" 2>/dev/null | awk -F/ 'NR==1{print $1}')"
+  [ -n "$rel_name" ] || die "could not read the release tarball"
+  if [ ! -d "$RELEASES/$rel_name" ]; then
+    tar -xzf "$art_file" -C "$RELEASES" 2>/dev/null || die "could not unpack $art_file"
+    good "unpacked $rel_name"
+  else
+    good "$rel_name already unpacked"
```
tella-butler, line 437: 🔴 **P1** — Retrying an interrupted extraction trusts a partial release

The extraction writes directly into the final release directory, while a retry treats any existing directory as complete. If the first run is killed, runs out of disk, or `tar` otherwise fails after creating `~/.opensession/releases/opensession-1.2.3-linux-x64`, the next run skips extraction, points `src` at that partial tree, and can install a shim whose executable or sidecars are missing. Match the update path: extract into a sibling staging directory, validate at least `release.json` and the expected executable, then atomically rename it into place. An existing destination without that completion validation must be removed and rebuilt.

scripts/lib/update.ts
```diff
@@ -26,14 +26,238 @@
...
+
+    // Reconcile the engine to the new release's pin BEFORE activating it: the
+    // engine, its plugin and the bundled SDK are one compatibility set, so a
+    // release that runs against a stale engine can fail every turn while
+    // update reports success. If a required component cannot be brought to the
+    // pin, leave the current release active (the src link is untouched) and
+    // return nonzero; the downloaded tree stays on disk for a later retry.
+    let reconciled = false;
+    try {
+      reconciled = await reconcileEngine(dest);
```
tella-butler, line 237: 🔴 **P1** — Failed engine reconciliation can still corrupt the active runtime

Moving reconciliation before the symlink swap prevents activation of the new app tree, but `reconcileEngine()` mutates the shared live engine first: it replaces the installed `opencode` binary and deletes `~/.config/opencode/{node_modules,package.json,package-lock.json}` before copying the new seed. For example, if the new engine installs successfully and the plugin copy then fails because the disk fills, this returns nonzero with the old release still active but a new engine and a missing or partial plugin runtime. The module's own compatibility contract says that combination can break every subsequent turn. Make reconciliation transactional by preserving the old engine and plugin state and restoring both on failure, or stage all replaceable state and commit it atomically only after every reconciliation step succeeds.

adrs/simple-mode.md
```diff
@@ -0,0 +1,348 @@
...
+## Definition of done (the "insanely simple" bar)
+
+On a fresh macOS or Ubuntu box with only `curl` and `git`:
+
+1. One command, no flags, no sudo prompt on the happy path.
+2. Zero questions before the server is running.
+3. It ends with a URL, and that URL opens a first-run page.
+4. First-run page: pick a display name, sign in a model account, pick or
+   create a repo. Three screens.
+5. First agent turn completes inside **five minutes** wall clock on a decent
```
tella-butler, line 45: 🔴 **P1** — The documented simple-mode result still includes a first turn that this PR removed

The current implementation intentionally removed `/setup`, yet this PR's definition of done still requires the installer URL to open that page, collect a model account, select a repository, and complete a first agent turn. On a fresh no-flag install without `OPENSESSION_CLAUDE_TOKEN`, onboarding creates the scratch repository and opens the regular UI, but there is no guided account flow or automatic first turn. The harness only exercises a real turn when an external token is supplied, so it does not establish the documented default path. Since the author has explicitly moved first-time usage to separate work, narrow this ADR, title, and PR description to the installation-and-running milestone, or restore the omitted flow before presenting this as reaching a first turn.

tella-butler (commented):
OS review · `4fa4d69`

tella-butler:
<details>

Not safe to merge. The latest commits fix the system-service migration ordering, but uninstall can still destroy preserved work, interrupted installs can activate incomplete releases, and failed engine reconciliation can break the active runtime. The documented simple-mode first-run result also remains unimplemented.
_5 inline comments below._
<sub>Reviewed `1904458` · Pi · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

</details>

install.sh
```diff
@@ -87,23 +122,151 @@ if [ "$DO_UNINSTALL" = "1" ]; then
...
+    # The install home holds the default worktrees dir and the scratch git repo
+    # its worktrees branch from; keep the whole thing if any of its worktrees is
+    # unsaved. A worktrees dir configured OUTSIDE the home is preserved on its
+    # own in the state sweep below.
+    if [ -n "$dirty_worktrees" ] && case "$worktrees_dir" in "$OPENSESSION_HOME"/*) true ;; *) false ;; esac; then
+      warn "keeping $OPENSESSION_HOME: session worktrees have unsaved work"
+      printf '%s' "$dirty_worktrees" | sed '/^$/d'
+      kept_any=1
+    else
+      rm -rf "$OPENSESSION_HOME"
```
tella-butler, line 238: 🔴 **P0** — Preserved external worktrees still lose their backing repository

The default release setup creates its scratch repository at `$OPENSESSION_HOME/scratch`, while `paths.worktreesDir` may be configured outside `$OPENSESSION_HOME`. If a dirty scratch worktree lives at `/srv/opensession-worktrees/scratch-feature`, the scan preserves that directory, but this branch deletes `$OPENSESSION_HOME` because the worktrees directory is external. The preserved worktree's `.git` file points into the now-deleted `$OPENSESSION_HOME/scratch/.git/worktrees/...`, so its uncommitted files remain but Git metadata, branch history, and unpushed commits become unusable. Preserve the backing repository for every dirty worktree. At minimum, keep `$OPENSESSION_HOME` whenever a dirty worktree is backed by the built-in scratch repository.

scripts/lib/update.ts
```diff
@@ -26,14 +26,238 @@
...
+    process.env.XDG_CONFIG_HOME || join(process.env.HOME || "~", ".config"),
+    "opencode",
+  );
+  if (existsSync(join(seedSrc, "node_modules"))) {
+    const want = seededPluginVersion(seedSrc);
+    const has = seededPluginVersion(ocConfig);
+    if (want && has !== want) {
+      try {
+        mkdirSync(ocConfig, { recursive: true });
+        for (const f of ["node_modules", "package.json", "package-lock.json"]) {
```
tella-butler, line 140: 🔴 **P1** — Failed engine reconciliation can corrupt the active runtime

Reconciliation mutates the global engine runtime before the release symlink is switched. It may first replace the `opencode` binary, then deletes the active plugin tree before copying the new seed. If that copy fails because the disk fills or permissions reject a file, the function returns false and keeps the old application release active, but that release is now paired with the new engine and a missing or partial plugin runtime. Existing turns can consequently fail even though update reports that it kept the current release. Stage the plugin tree and swap it atomically, and restore both the previous plugin runtime and engine version on any later reconciliation failure.

adrs/simple-mode.md
```diff
@@ -0,0 +1,348 @@
...
+reinstalling: same files, same paths, more of them filled in.
+
+## Definition of done (the "insanely simple" bar)
+
+On a fresh macOS or Ubuntu box with only `curl` and `git`:
+
+1. One command, no flags, no sudo prompt on the happy path.
+2. Zero questions before the server is running.
+3. It ends with a URL, and that URL opens a first-run page.
+4. First-run page: pick a display name, sign in a model account, pick or
```
tella-butler, line 43: 🔴 **P1** — The stated simple-mode result still depends on a removed first-run flow

The definition of done says the installed URL opens a three-screen first-run page and completes a real first turn, but this PR removed `/setup` and now opens the regular application. On a fresh box without `OPENSESSION_CLAUDE_TOKEN`, the one-command path therefore ends without a configured model account or the promised guided first turn. Either restore the setup flow required by this ADR and the PR's stated goal, or explicitly scope this change to installation and remove the first-run and first-turn claims from the PR and its acceptance document.

tella-butler (commented):
OS review · `1904458`

tella-butler:
Not safe to merge. The current installer can still destroy Git metadata for preserved worktrees and trust an interrupted release extraction. Release updates can also damage the active engine before refusing activation, while the documented first-turn flow remains absent. The system-service migration issue is addressed, but macOS paths are again unquoted at the current HEAD.
_5 inline comments below._
<sub>Reviewed `391cdb0` · Pi · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-100-review)</sub>

PR #101, opened 2026-08-18
Centralize session ownership and extract the executor

jfrolich:
## Summary
- move writable SessionKernel state and per-session lease coordination into a versioned Worker actor
- keep the detached executor independently restartable while making lifecycle decisions belong to one session owner
- route REST, WebSocket, MCP, automation, queue, ask, transcript, session-file, timer, and executor mutations through durable commands
- retain permanent compact request fingerprints, generation-fenced run state, durable timers/effects, permanent deletion tombstones, and deterministic create plans

## Command and retry semantics
- capable web/native clients retain every unresolved mutation and replay the same request id after reconnect or restart
- completed request fingerprints are permanent; large semantic results remain fully replayable until a durable client/server acknowledgement handshake completes, then compact after 30 days
- the web outbox migrates the shipped aggregate key in place and stores each request separately, with finite legacy tombstones preventing partial-migration and cross-tab resurrection
- terminal failures are sticky and bounded; retryable failures can re-enter only when the server call site explicitly declares the command replay-safe
- actor loss makes non-replay-safe physical work `indeterminate`; known create, prompt, cancel, answer, queue, and timer paths use stable adoption identities
- create ownership is scoped to the deterministic target session and releases once the session plus opening dispatch are durable, so one slow create cannot block another

## Recovery and effects
- stale host hello, transcript, ask, steer, and completion frames are generation-fenced
- sandbox launch phases and host-produced evidence distinguish safe replay from uncertain execution
- ambiguous launches retain a live reconciliation owner, remain cancellable, and never infer absence from missing connectivity
- command completion and outbox staging commit together; destination delivery is documented as idempotent at-least-once
- timers and effects back off and dead-letter; unknown version-skew kinds cannot head-of-line block registered work
- workspace admins can paginate, retry, or discard dead letters, with mutation audit records

## Deployment and security
- the executor fails closed when configured but unavailable; direct launch requires the explicit `OPENSESSION_EXECUTOR=0` operator bypass
- the executor unit receives only path settings and systemd credentials, not the application secrets EnvironmentFile
- run hosts still receive the gateway EnvironmentFile through the fixed root-owned helper
- schema migration acquires the writer claim first; tracked schema floor 6 blocks unsafe automatic rollback to an older reader
- projected GitHub credentials are removed on startup validation failures, cancellation, and normal exit

## Verification
- final head `8ecdcc568923a5c91311eca99691261888a13474` is rebased onto current `main` and merge-tree clean
- `bun run typecheck` passes
- strict transcript snapshots pass, 6 of 6
- focused actor, executor, create, queue, ask, client-outbox, deploy, host, sandbox, and GitHub suites pass
- 371 server, executor, and runner-host modules import without listener, timer, or subprocess side effects
- isolated dev boot reports a healthy, non-degraded actor and the expected executor bypass state
- all GitHub CI checks pass, including the new gating ownership/executor suite
- independent exact-head adversarial review reports no remaining high or medium issue

The native CI gate builds both iOS and macOS schemes and runs the macOS test suite on every relevant pull request. It passes on the exact final head.


tella-butler:
<details>

```mermaid
sequenceDiagram
  participant K as Session kernel
  participant E as Executor
  participant S as systemd
  participant H as Run host
  K->>E: launch_host(hostId, specHash)
  E->>E: persist starting
  E->>S: launch fixed unit
  S->>H: start persisted spec
  E--xK: response lost
  H->>H: execution evidence written
  K->>E: retry launch_host
  E->>S: relaunch if unit is now inactive
```

</details>

packages/core/opensession-server/src/executor/coordinator.ts
```diff
@@ -0,0 +1,507 @@
...
+    if (existing && existing.specHash !== specHash) {
+      throw new CoordinatorError(
+        "spec_hash_mismatch",
+        "host id is already bound to another run spec",
+      );
+    }
+    if (existing?.state === "uncertain") {
+      return this.reconcileUncertain(hostId, specHash, existing.error);
+    }
+    if (existing && existing.state !== "starting") {
```
tella-butler: 🔴 **P1** — Recovered starting records can replay an observed run

A persisted `starting` record falls through to a fresh `launchOnce`, and after an executor replacement there is no in-memory `inflight` entry. This path never checks `hostStarted` before `launchOnce` calls `deps.launch`. For example, the executor can persist `starting`, successfully start `rh-X`, then die before recording `started`; the host writes `meta.json` and performs a tool side effect, but also dies while the executor remains unavailable. A later reconciliation request sees the unit inactive and launches the original spec again, replaying the turn despite durable execution evidence. Before relaunching a recovered `starting` record, probe readiness, unit state, and `hostStarted`; return `launch_uncertain` when metadata proves the prior host started, and relaunch only when both the effect and execution evidence are proven absent.

packages/core/opensession-server/src/server/host-client.ts
```diff
@@ -300,27 +317,55 @@ async function spawnHostRun(
...
+  } catch (cause) {
+    let error = cause;
+    if (launchCompleted) {
+      try {
+        await stopAndVerifyHostAbsent(hostId, dir);
+      } catch (cleanupError) {
+        error = cleanupError;
+      }
+    }
+    if (!(error instanceof ExecutorProtocolError && error.ambiguousLaunch)) {
```
tella-butler, line 408: 🔴 **P1** — Ambiguous launches cannot be cancelled while disconnected

Keeping the handle registered avoids falsely declaring an uncertain launch dead, but its `cancel` control only sends over `this.conn`. If the executor outcome is ambiguous and the host remains unconnectable for the 120-second wait, the handle is retained with no connection. The UI Stop path then calls `cancelAgentRun`: `hostCancel` returns false, the journal is cleared, but no `systemctl stop` or executor `stop_host` occurs and the handle is never unregistered. The session remains busy indefinitely, while the orphaned unit may later start and execute a turn whose recovery record was deleted. Give uncertain local handles a launcher-level stop-and-verify cancellation path, and clear the journal/control only after host absence is proven; if absence remains uncertain, retain ownership and report that Stop could not be confirmed.

tella-butler (commented):
OS review · `8b4109d`

tella-butler:
<details>

```mermaid
sequenceDiagram
  participant K as Session kernel
  participant E as Executor
  participant H as Run host
  K->>E: launch_host
  E->>E: persist starting
  E->>H: launch
  H->>H: begin turn
  E--xK: response lost / executor restarts
  K->>E: retry launch_host
  E->>H: relaunch persisted spec
  Note over H: Original prompt can execute twice
```

</details>

tella-butler:
<details>

```mermaid
sequenceDiagram
  participant K as Session kernel
  participant E as Executor
  participant H as Run host
  K->>E: launch_host
  E->>E: persist starting
  E->>H: launch unit
  H->>H: write meta and run prompt
  E--xK: disconnect or restart
  K->>E: retry launch_host
  alt host exited before retry
    E->>H: launch same prompt again
  else host socket unavailable
    K->>K: register disconnected control
    K--xH: Stop is dropped
  end
```

</details>

tella-butler (commented):
OS review · `089f820`

packages/core/opensession-server/src/executor/coordinator.ts
```diff
@@ -0,0 +1,507 @@
...
+  ): Promise<ExecutorHostStatus> {
+    const dir = `${this.hostsDir}/${hostId}`;
+    if (await this.deps.hostReady(dir)) {
+      this.recordStarted(hostId, specHash);
+      return this.status(hostId);
+    }
+
+    try {
+      if (!(await this.deps.unitActive(hostId))) {
+        await this.deps.launch(hostId, dir, specHash);
```
tella-butler, line 358: 🔴 **P1** — Recovered starting records can replay an observed run

A persisted `starting` record is treated like a fresh launch: after an executor replacement there is no in-memory `inflight` entry, so `launchOnce` launches again whenever the old unit is no longer active. The run host writes `meta.json` before invoking `runAgent`, which gives a concrete failure window: the first host writes metadata, executes a short prompt with tool side effects, and exits while the executor is down; the retry sees `hostReady=false` and `unitActive=false` and starts the identical spec again. That duplicates the turn and can duplicate messages, commits, or other tool effects, contrary to the documented idempotency guarantee. When recovering an existing `starting` record, check `hostReady`, `unitActive`, and `hostStarted` before launching. If metadata proves a host started, preserve an uncertain/recoverable state instead of replaying; launch only when both the unit and execution evidence are absent.

packages/core/opensession-server/src/server/host-client.ts
```diff
@@ -300,27 +317,55 @@ async function spawnHostRun(
...
+    } catch (error) {
+      if (!(error instanceof ExecutorProtocolError && error.ambiguousLaunch)) {
+        throw error;
+      }
+      handle = new HostHandle(dir, spec, {
+        onAskUser: opts.onAskUser,
+        onSteerFailed: opts.onSteerFailed,
+      });
+      try {
+        await handle.connectWithWait(120_000);
```
tella-butler, line 386: 🔴 **P1** — Ambiguous launches cannot be cancelled while disconnected

This path registers a `HostHandle` and waits up to 120 seconds for a socket, but its registered cancel control only calls `send({ t: "cancel" })`, which returns false while disconnected. A user pressing Stop during this window therefore only latches the preparation token; `runAgentHosted` receives no `startToken` and performs no cancellation check after connection, so an already-running detached prompt continues and is accepted once the socket appears. The executor protocol already defines `stop_host`, but the client never sends it. Make cancellation during launch/connect durable: abort the wait and stop the host through the executor with verified direct cleanup as fallback, then clear the handle and journal only after absence is proven.

tella-butler:
<details>

```mermaid
sequenceDiagram
  participant K as Session kernel
  participant E as Executor
  participant H as Run host
  K->>E: launch_host(hostId, specHash)
  E->>E: persist starting
  E->>H: launch unit
  H->>H: write meta.json, run prompt
  E--xK: disconnect / executor restart
  K->>E: retry launch_host
  alt host already exited
    E->>H: launch same spec again
  else socket not up yet
    K--xH: Stop is dropped
  end
```

</details>

tella-butler:
<details>

Not safe to merge. The two previously reported duplicate-execution and cancellation gaps remain, and the latest GitHub recovery change clears the journal for a launch it explicitly considers uncertain. These paths can run the same turn twice despite the PR's central safety guarantee.
_3 inline comments below._
<sub>Reviewed `b56c9fb` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

packages/core/opensession-server/src/agents/github/run.ts
```diff
@@ -424,6 +424,11 @@ export async function runGithubAgent(opts: GithubRunOpts): Promise<GithubRunResu
       recoveredRun = journalStartRecovery(recoveredRun);
       engineSessionId = recoveredRun.claudeSessionId || engineSessionId;
       const reattached = await resumeLocalHostRun(recoveredRun, {});
+      if (reattached === "uncertain") {
```
tella-butler, line 440: 🔴 **P1** — GitHub recovery discards the journal for an uncertain host

Throwing here does not preserve the uncertainty: the surrounding catch converts it into `errorMsg`, and the unconditional `finally` calls `journalClearIfLineage(recoveredRun)`. For a same-head review whose host is alive but temporarily not connectable, this deletes the only recovery marker. The next automation retry sees no detached run and can launch a second review while the first host is still executing, potentially duplicating review comments or code changes. Track this outcome separately and leave the journal claimed until the host is reattached or proven dead.

tella-butler (commented):
OS review · `b56c9fb`

tella-butler:
<details>

Safe once the remaining P1 recovery gap is fixed. The PR centralizes session ownership and moves detached launches into the executor, and the two earlier executor replay and cancellation issues are addressed. However, uncertain GitHub host recovery still destroys the durable state needed for a later reattach.
_1 inline comment below._
<sub>Reviewed `5fea85e` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

tella-butler:
<details>

Safe once the P1 below is fixed. The latest commits only merge changes from main, so the executor and session-ownership implementation is unchanged since the previous review. GitHub recovery still abandons an active but temporarily unreachable detached run.
_1 inline comment below._
<sub>Reviewed `adfd542` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

tella-butler:
<details>

Safe once the P1 below is fixed. The executor and session-kernel ownership split now avoids the previously reported launch replay paths, but GitHub recovery still loses ownership when a detached host is unreachable yet potentially alive.
_1 inline comment below._
<sub>Reviewed `d67af81` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

tella-butler:
<details>

Safe once the P1 below is fixed. The latest commits do not address the remaining recovery blocker. An uncertain GitHub run can still lose its durable journal entry while its detached host continues executing, defeating restart recovery and allowing a later retry to overlap it.
_1 inline comment below._
<sub>Reviewed `169a6dd` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

tella-butler:
<details>

### 🤖 OS review

</details>

tella-butler:
<details>

Safe once the P1 below is fixed. The prior uncertain GitHub recovery issue is addressed, and the actor/executor ownership fencing now holds together. One completion failure path still violates the stated fail-stop guarantee and can replay an already-applied command.
_1 inline comment below._
<sub>Reviewed `5105f42` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

packages/core/opensession-server/src/server/session-kernel/actor-worker.ts
```diff
@@ -0,0 +1,254 @@
...
+            effects: request.effects,
+          });
+        post({ t: "complete_result", rpcId: request.rpcId });
+      } else {
+        store.failCommand(lease.sessionId, lease.requestId, request.error);
+        post({ t: "fail_result", rpcId: request.rpcId });
+      }
+      active.delete(lease.sessionId);
+      grantNext(lease.sessionId);
+    } catch (error) {
```
tella-butler, line 172: 🔴 **P1** — Receipt commit failures remain retryable after the command effect ran

If a command handler has already applied its gateway effect and `completeCommandDecision` then throws, for example with `SQLITE_FULL` or `SQLITE_IOERR`, this catch reports an ordinary RPC error while leaving the actor alive. `SessionKernel.dispatch` consequently calls `fail`, which either leaves this session's active lease permanently wedged if persistence still fails, or records the command as failed and allows the same request ID to execute its already-applied effect again. Other sessions continue working, so the gateway looks healthy despite losing the command's ownership guarantee. Treat completion persistence errors as ambiguous fatal actor failures: terminate the actor so the client triggers gateway fail-stop, and do not convert a failed completion commit into a retryable `failCommand`.

tella-butler (commented):
OS review · `5105f42`

tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `b24adf0`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)

</details>

tella-butler:
<details>

### 🤖 OS review

</details>

tella-butler:
<details>

Safe once the two P1 ownership gaps below are fixed. The new commits correctly preserve uncertain GitHub recovery and improve sandbox replay fencing, but command settlement can still make completed effects retryable, and uncertain sandbox launches cannot be cancelled while disconnected.
_2 inline comments below._
<sub>Reviewed `604efb3` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

packages/core/opensession-server/src/server/session-kernel/actor-worker.ts
```diff
@@ -0,0 +1,254 @@
...
+      if (request.t === "complete") {
+        if (!store.isTombstoned(lease.sessionId))
+          store.completeCommandDecision({
+            sessionId: lease.sessionId,
+            requestId: lease.requestId,
+            type: lease.type,
+            result: request.result,
+            effects: request.effects,
+          });
+        post({ t: "complete_result", rpcId: request.rpcId });
```
tella-butler, line 160: 🔴 **P1** — Receipt commit failures remain retryable after the handler ran

If `completeCommandDecision` throws after the gateway handler has already performed its mutation, `finish` returns an ordinary RPC error while retaining the lease. `SessionKernel.dispatch` then calls `actor.fail`, which unconditionally changes the receipt to `failed` and releases the lease. A client replaying the same request ID is therefore admitted again because `grant` deduplicates only `completed` records, so the handler can run twice. Completion uncertainty must not be converted into a normal command failure. Fail-stop the actor/gateway and preserve an uncertain receipt, then reconcile that receipt without rerunning the handler unless its effects are proven safe to replay.

packages/core/opensession-server/src/server/sandbox/docker.ts
```diff
@@ -1139,34 +1150,64 @@ function makeDockerSandbox(
...
-      const gen = withRunJournal(handle.events(), record);
+      const ownedHandle = handle!;
+      const rawEvents = uncertainLaunch
+        ? (async function* (): AsyncGenerator<StreamEvent> {
+            for (;;) {
+              try {
+                await ownedHandle.connectWithWait(60_000);
+                break;
+              } catch (error) {
+                console.warn(`[sandbox] ${spec.hostId}: still waiting for uncertain host`, error);
```
tella-butler: 🔴 **P1** — Uncertain sandbox launches cannot be cancelled

After `docker exec -d` is admitted but the host does not connect within 30 seconds, this loop retains ownership and retries forever. However, both the Docker launcher and the analogous remote launcher omit `HostLauncher.stop`. While disconnected, `HostHandle.cancelHost()` cannot send a cancel frame and therefore returns false. For example, if the detached command was accepted but the host crashes before dial-back, pressing Stop cannot terminate this loop: the host remains registered as busy, the journal remains owned, and queued prompts cannot drain. Implement a verified `stop` operation for both sandbox launchers, or otherwise make this reconciliation owner observe cancellation and prove the launched process absent before releasing ownership.

tella-butler (commented):
OS review · `604efb3`

tella-butler:
<details>

Safe once the two P1s below are fixed. The GitHub uncertain-host recovery issue is now addressed, but the new sandbox cancellation path cannot reliably prove termination, and command settlement failures can still make an already-executed handler retryable.
_2 inline comments below._
<sub>Reviewed `77edb47` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

packages/core/opensession-server/src/server/sandbox/docker.ts
```diff
@@ -985,12 +993,29 @@ function makeDockerLauncher(container: string, sessionId: string): HostLauncher
...
+        throw new HostLaunchNotDispatchedError(
+          `docker exec (run host) failed: ${r.stderr.trim().slice(0, 400)}`,
+        );
       }
     },
+    async stop(hostId, dir) {
+      await Bun.write(`${dir}/cancelled`, "cancelled\n");
+      const specPath = assertSafePath(`${dir}/${HOST_SPEC_NAME}`);
+      const script =
+        `pids="$(pgrep -f '${specPath}' 2>/dev/null || true)"; ` +
```
tella-butler: 🔴 **P1** — The cancellation probe matches and kills its own shell

`pgrep -f` searches the full command line, and the parent `sh -c` command line itself contains `specPath`. On an ambiguous Docker launch, this probe therefore includes the cleanup shell's PID in `pids`; the subsequent `kill -TERM` terminates the cleanup before it reaches the absence check. `stopAndWait()` then reports failure and `reconcileUncertainHostEvents()` retains recovery ownership indefinitely, even when the host was stopped or never started. The identical problem exists in `sandbox/adapters/bootstrap.ts:1661`. Use the PID recorded in the host metadata and verify that process, or construct a self-excluding search pattern, rather than matching the literal path embedded in the cleanup command.

packages/core/opensession-server/src/server/session-kernel/actor-worker.ts
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  function finish(
+    request: Extract<KernelActorAsyncRequest, { t: "complete" | "fail" }>,
+  ): void {
+    try {
+      const lease = leaseFor(request.leaseId);
+      if (request.t === "complete") {
+        if (!store.isTombstoned(lease.sessionId))
+          store.completeCommandDecision({
```
tella-butler, line 153: 🔴 **P1** — Receipt commit failures remain retryable after the handler ran

The gateway runs the command handler before sending `complete`. If `completeCommandDecision()` throws, `finish()` returns an error while retaining the lease; `SessionKernel.dispatch()` catches that settlement error and calls `actor.fail()`, which marks the receipt `failed` and releases the lease. A client replaying the same request ID is then admitted again because `acceptCommand()` only deduplicates `completed` records, so the handler can execute twice after its first physical mutation already succeeded. Separate handler failure from completion failure: once the handler returns, do not convert a failed receipt commit into `failed`; retain/fail-stop ownership or retry the completion decision until its outcome is known.

tella-butler (commented):
OS review · `77edb47`

tella-butler:
<details>

### 🤖 OS review

</details>

tella-butler:
<details>

Safe once the P1 below is fixed. The latest changes correctly preserve uncertain GitHub recovery records and close the sandbox cancellation races, but command effects can still occur without a durable receipt, breaking the PR's restart-idempotency guarantee.
_1 inline comment below._
<sub>Reviewed `dc29617` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `bbf1b72`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)

</details>

tella-butler:
<details>

Safe once the remaining P1 is fixed. The PR establishes the actor-owned session mailbox and independently recoverable executor, and the uncertain GitHub-host and sandbox cancellation paths now preserve ownership correctly. Receipt settlement can still convert an already-executed command into a retryable failure.
_1 inline comment below._
<sub>Reviewed `d8349b9` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

packages/core/opensession-server/src/server/session-kernel/actor-worker.ts
```diff
@@ -0,0 +1,254 @@
...
+            sessionId: lease.sessionId,
+            requestId: lease.requestId,
+            type: lease.type,
+            result: request.result,
+            effects: request.effects,
+          });
+        post({ t: "complete_result", rpcId: request.rpcId });
+      } else {
+        store.failCommand(lease.sessionId, lease.requestId, request.error);
```
tella-butler: 🔴 **P1** — Receipt commit failures remain retryable after the handler ran

`SessionKernel.dispatch` uses one catch block for both handler failures and `actor.complete(...)` failures. If the handler succeeds but `completeCommandDecision` fails, for example because SQLite is temporarily locked, the client immediately sends `fail` for the same lease. This line persists the receipt as `failed` and releases the mailbox. A replay of that request is then admitted again because `acceptCommand` deduplicates only `completed` records, so the handler runs twice. A concrete case is `answerQuestion`: the first execution resolves the pending ask, receipt completion fails, and replay executes after the ask has already been consumed and returns `false`, despite the answer having been delivered. Other handlers can repeat physical mutations or launches. Separate handler failure from settlement failure: once the handler returns, never mark that command failed. Retry the completion decision under the existing lease, or fail-stop while retaining an indeterminate/processing receipt that requires command-specific recovery before replay.

tella-butler (commented):
OS review · `d8349b9`

tella-butler:
<details>

Safe once the P1 below is fixed. The latest changes address the previously reported uncertain GitHub recovery, sandbox cancellation, and command receipt settlement failures. The remaining writer-ownership check can reject the only valid actor after ordinary PID reuse, preventing the gateway from starting.
_1 inline comment below._
<sub>Reviewed `989dd94` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

packages/core/opensession-server/src/server/session-kernel/store.ts
```diff
@@ -0,0 +1,1012 @@
...
+
+	private claimWriter(): void {
+		const transaction = this.db.transaction(() => {
+			const current = this.db
+				.query("SELECT owner_id, pid FROM session_kernel_owner WHERE singleton = 1",)
+				.get() as { owner_id: string; pid: number } | null;
+			if (current && current.owner_id !== PROCESS_OWNER_ID) {
+				let alive = false;
+				try {
+					process.kill(current.pid, 0);
```
tella-butler, line 458: 🔴 **P1** — A reused PID can prevent the session actor from starting

`process.kill(current.pid, 0)` proves only that some process currently has this numeric PID, not that it is the process which wrote `owner_id`. For example, after Open Session is stopped long enough for its old PID 2400 to be assigned to an unrelated long-lived process, every new actor sees that process as alive and throws here even though the recorded owner is gone. Because actor startup gates the gateway boot, the service remains unavailable until that unrelated process exits or the database row is manually removed. Persist and compare a process-birth identifier such as `/proc/<pid>/stat` start time alongside the PID, or use an OS-held lock whose ownership disappears when the process dies.

tella-butler (commented):
OS review · `989dd94`

tella-butler:
<details>

### 🤖 OS review

</details>

tella-butler:
<details>

### 🤖 OS review

</details>

tella-butler:
<details>

Safe once the two P1 ownership failures below are fixed. The recovery changes now preserve uncertain GitHub journals and fail-stop ambiguous successful completions, but the actor can still wedge a session after a failed receipt write, and stale PID reuse can prevent the service from booting.
_2 inline comments below._
<sub>Reviewed `ece2b4f` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

packages/core/opensession-server/src/server/session-kernel/actor-worker.ts
```diff
@@ -0,0 +1,299 @@
...
+          store.completeCommandDecision({
+            sessionId: lease.sessionId,
+            requestId: lease.requestId,
+            type: lease.type,
+            result: request.result,
+            effects: request.effects,
+          });
+        post({ t: "complete_result", rpcId: request.rpcId });
+      } else {
+        store.failCommand(
```
tella-butler, line 162: 🔴 **P1** — A failed retry-receipt commit leaves the session mailbox permanently leased

If a replay-safe handler fails and `store.failCommand` then throws, for example because SQLite reports `SQLITE_FULL` or an I/O error, `finish` only posts an error. It never removes the lease from `active`, and `SessionKernelActorClient.fail` does not mark the actor dead. The original timer command rejects, but every later command for that session remains queued behind the orphaned lease while the gateway continues serving. Treat any settlement persistence failure as fatal, as `complete` now does: terminate the actor and fail-stop the gateway, or otherwise prove the receipt committed before releasing the mailbox.

tella-butler (commented):
OS review · `ece2b4f`

tella-butler:
<details>

Safe once the two P1 failure modes below are fixed. The actor and executor split otherwise holds together, including the latest transcript ordering changes. I retract the prior uncertain GitHub recovery and post-handler completion findings: the current code preserves uncertain journals and fail-stops ambiguous completion. A stale PID can still prevent startup, and a failed retry-receipt commit can permanently block one session mailbox.
_2 inline comments below._
<sub>Reviewed `6aabe56` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

tella-butler:
<details>

Safe once the three P1 durability gaps below are fixed. The new commits address the earlier receipt-commit and uncertain-host recovery blockers, but commands can still be lost or reassigned during WebSocket negotiation, and schema-v4 ownership records can still prevent startup after PID reuse.
_3 inline comments below._
<sub>Reviewed `a8443cc` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

packages/core/opensession-server/src/frontend/hooks/useWebSocket.ts
```diff
@@ -363,8 +440,45 @@ export function useWebSocket(presenceActive = true) {
...
+
   const send = useCallback(
     (msg: WSClientMessage) => {
+      msg = withMutationRequestId(msg);
+      const mutationRequestId =
+        "requestId" in msg && typeof msg.requestId === "string"
+          ? msg.requestId
+          : undefined;
+      if (mutationRequestId && !commandNegotiatedRef.current) {
+        negotiatingCommandsRef.current.set(mutationRequestId, msg);
```
tella-butler, line 514: 🔴 **P1** — Commands accepted before negotiation are not durable

Before the server hello arrives, this records the mutation only in a React ref and returns successfully. For example, if someone submits a prompt while the socket is reconnecting and then the PWA is killed or the page reloads before `finishCommandNegotiation`, the ref disappears, nothing was written to localStorage, and the server never received the prompt. This violates the PR's guarantee that every unresolved mutation survives reconnects and restarts. Persist commands in a pre-negotiation partition before returning, then atomically promote them to the verified `commandScope`, or reject the mutation if durable staging fails.

packages/core/opensession-server/src/frontend/hooks/useWebSocket.ts
```diff
@@ -363,8 +440,45 @@ export function useWebSocket(presenceActive = true) {
     syncPresenceRef.current();
   }, [presenceActive]);
 
+
+  useEffect(() => {
+    const reconnectForIdentity = () => {
+      commandNegotiatedRef.current = false;
+      commandResultsRef.current = false;
+      wsRef.current?.close();
```
tella-butler: 🔴 **P1** — Identity changes carry queued commands into the new scope

The identity-change handler resets negotiation and closes the socket but leaves `negotiatingCommandsRef` intact. In local multi-user mode, if Ada queues a prompt while disconnected and then switches to Grace before reconnection, Grace's hello causes `finishCommandNegotiation` to persist and send Ada's prompt under Grace's scope and attribution. Partition the negotiating queue by the identity that created each command, and never promote entries from the previous identity into the new verified scope.

packages/core/opensession-server/src/server/session-kernel/store.ts
```diff
@@ -0,0 +1,1180 @@
...
+					alive = true;
+				} catch {}
+				if (alive) {
+					const recorded = parseOwnerIdentity(current.owner_id);
+					const bootId = linuxBootId();
+					const start = linuxProcessStart(current.pid);
+					if (
+						recorded?.bootId && recorded.start &&
```
tella-butler, line 466: 🔴 **P1** — Legacy ownership records still fail closed on a reused PID

Schema-v4 databases contain a plain UUID in `owner_id`, so `parseOwnerIdentity` returns undefined and this new identity comparison is skipped. If the old gateway stopped while owning PID 4242 and Linux later assigned 4242 to an unrelated live process, every v5 actor startup treats that process as the old owner and throws `Session kernel already owned` until it exits or the database is repaired manually. Handle legacy owner rows explicitly, for example by comparing the current process start time with the recorded `claimed_at` during the v4-to-v5 migration, so PID reuse can be distinguished without weakening live-writer exclusion.

tella-butler (commented):
OS review · `a8443cc`

tella-butler:
The new admin-only Reliability panel exposes failed timer and external deliveries without requiring raw API calls. It is responsive at both desktop and phone widths.

Desktop:
Reliability panel at desktop width

Phone:
Reliability panel at phone width

tella-butler:
<details>

Not safe to merge. The ownership and executor split is substantially hardened, but the new readiness gate never becomes ready, identity changes can replay another user's queued mutation, and one legacy ownership case can prevent startup. I am dropping the earlier GitHub-journal and generic receipt-settlement findings because the current uncertainty preservation and fail-stop behavior cover those cases.
_3 inline comments below._
<sub>Reviewed `940b8b8` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

packages/core/opensession-server/src/server/routes/system.ts
```diff
@@ -18,12 +18,103 @@ import { getSessionControl } from "../session-control";
...
+	if (path === "/live" && req.method === "GET")
+		return Response.json({ ok: true, bootId: BOOT_ID, uptime: process.uptime() });
+
+	if (path === "/ready" && req.method === "GET") {
+		try {
+			const kernel = await sessionKernelHealth();
+			const executor = executorClientHealth();
+			const executorReadiness = await executorClientReady();
+			const readiness = serviceReadiness();
+			const ready = executorReadiness.ready && readiness.phase === "ready";
```
tella-butler, line 105: 🔴 **P1** — The readiness endpoint can never report ready

`service-readiness.ts` initializes the phase to `booting`, but the only production calls to `setServiceReadiness` set `failed` or `draining`; nothing sets `recovering` or `ready`. Consequently every healthy cold boot returns 503 from `/ready`, causing `deploy.sh` and `self-deploy.sh` to fail their health gates and potentially attempt rollback. Set `recovering` before recovery begins and set `ready` only after reconciliation and `startSessionKernelRuntime()` complete.

packages/core/opensession-server/src/frontend/hooks/useWebSocket.ts
```diff
@@ -96,10 +106,57 @@ export function useWebSocket(presenceActive = true) {
...
+            if ("requestId" in candidate && typeof candidate.requestId === "string")
+              negotiatingCommandsRef.current.set(candidate.requestId, candidate);
+            commandResultsRef.current = false;
+            const failedSocket = wsRef.current;
+            wsRef.current = null;
+            failedSocket?.close();
+            setTimeout(() => connect(), 250);
+            return;
+          }
+          const provisional = wsCommandOutboxForScope(localCommandScope());
```
tella-butler: 🔴 **P1** — Negotiating commands still cross identity boundaries

The identity-change handler reconnects but leaves `negotiatingCommandsRef` intact. If Ada issues a mutation before `hello`, then switches to Grace, the new socket takes Ada's candidate from that map, persists it into Grace's server-provided command scope, and sends it as Grace. This also leaves Ada's provisional record behind because `localCommandScope()` is re-evaluated after the switch. Track the provisional scope with each candidate, and on identity change either partition the in-memory queue by scope or clear it while leaving the old scoped durable record for Ada's next connection.

packages/core/opensession-server/src/server/session-kernel/store.ts
```diff
@@ -0,0 +1,1226 @@
...
+						bootId && start &&
+						(recorded.bootId !== bootId || recorded.start !== start)
+					) alive = false;
+					else if (!recorded && !plausibleLegacyOwner(current.pid)) alive = false;
```
tella-butler, line 470: 🔴 **P1** — A legacy owner row can reject its own replacement process

The new command-line check fixes reuse by unrelated processes, but a stale legacy row whose PID is reused by the replacement server still passes `plausibleLegacyOwner`: `/proc/<pid>/cmdline` is now this process and contains `opensession.ts`. Because the legacy row has no start identity, `claimWriter()` then treats the current process as a competing live owner and aborts every startup with that PID. Treat `current.pid === process.pid` as stale when `owner_id` is not the current process identity, before applying the legacy command-line fallback.

tella-butler (commented):
OS review · `940b8b8`

tella-butler:
<details>

Safe once the two P1 identity and ownership races below are fixed. The new fail-stop settlement behavior and production readiness transition address the prior receipt and readiness blockers, but queued WebSocket mutations can still cross users and a reused legacy PID can still prevent the authoritative actor from starting.
_2 inline comments below._
<sub>Reviewed `9d856e2` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

</details>

packages/core/opensession-server/src/frontend/hooks/useWebSocket.ts
```diff
@@ -96,10 +106,64 @@ export function useWebSocket(presenceActive = true) {
...
+      for (const ack of commandOutbox.pendingAcks()) {
+        try { ws.send(JSON.stringify(ack)); } catch {}
+      }
+      const existing = commandOutbox.pending();
+      const existingIds = new Set(existing.map((command) => command.requestId));
+      for (const command of existing) {
+        try { ws.send(JSON.stringify(command)); } catch {}
+      }
+      const candidates = new Map<string, WSClientMessage>();
+      for (const candidate of [...provisional.pending(), ...inMemory])
```
tella-butler, line 149: 🔴 **P1** — Pre-negotiation commands still cross identity boundaries

The identity-change handler reconnects without clearing `negotiatingCommandsRef`, and this line merges that unscoped in-memory map into whichever verified scope the next socket announces. For example, while the server is unavailable Alice submits a prompt, then the browser switches to Bob before negotiation completes. Bob's eventual `hello` advertises `github:bob`, but Alice's command remains in `inMemory`, is persisted into Bob's outbox, and is sent under Bob's authenticated socket. Clear the in-memory map on identity change, leaving its already durable command in Alice's provisional outbox, or store the originating local scope with each entry and migrate only matching entries.

tella-butler (commented):
OS review · `9d856e2`

tella-butler:
<details>

### 🤖 OS review

</details>

tella-butler:
Not safe to merge until the two P1 issues below are fixed. The actor and executor split now handles readiness and durable pre-negotiation storage, but legacy ownership recovery can still block startup and in-memory negotiating commands can cross user identities. I am retracting the earlier receipt-settlement findings because the fail-stop path and replay-safe startup reconciliation cover those cases.
_2 inline comments below._
<sub>Reviewed `3f2a4b3` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-101-review)</sub>

packages/core/opensession-server/src/frontend/hooks/useWebSocket.ts
```diff
@@ -363,8 +477,53 @@ export function useWebSocket(presenceActive = true) {
     syncPresenceRef.current();
   }, [presenceActive]);
 
+
+  useEffect(() => {
+    const reconnectForIdentity = () => {
+      commandNegotiatedRef.current = false;
```
tella-butler, line 483: 🔴 **P1** — Pre-negotiation commands still cross identity boundaries

The identity-change handler resets negotiation and replaces the socket but leaves `negotiatingCommandsRef` intact. If Alice sends `create_session` before the hello arrives and then signs out, the command remains in this map. After Bob signs in, Bob's hello calls `finishCommandNegotiation`, which takes that in-memory command, writes it into Bob's verified outbox, and sends it over Bob's authenticated socket. This executes Alice's intent in Bob's scope despite the durable provisional copy being correctly partitioned. Clear the negotiating map when identity changes; Alice's command remains safely stored in Alice's provisional outbox and can replay when Alice reconnects.

tella-butler (commented):
OS review · `3f2a4b3`

PR #102, opened 2026-08-19
A user's "new session" means a real session, not a subagent

happylinks:
When someone says "create a new session for X", they mean a detached Open Session session created as them: it lands in their own sidebar, has its own worktree/branch and transcript, and keeps running after the current turn ends. Models were reading that as an invitation to spin up an in-process subagent instead — invisible in the sidebar, impossible for the user to open or steer, and gone when the run finishes.

This writes the distinction into the product rather than leaving it to per-user memory:

- `run-instructions.ts` — the `Managing` block (gated on the sessions MCP actually being mounted, so runs without it hear nothing) now says a user's "a new session" means `create_session`, never a subagent, and that the reply should carry the new session's URL. Subagents and `spawn_task` stay what they are: work the agent chooses to fan out inside its own turn, or an explicit user request for a subagent/worker. The ambiguous case defaults to creating the session — one they didn't need is easy to close, a subagent they wanted to open doesn't exist.
- `create_session`'s own tool description says the same thing at the point of use.
- Two tests in `zz-opencode.test.ts` lock both directions: present with the sessions server, absent without it.

`bun test src/server/zz-opencode.test.ts` → 96 pass, and `bun run typecheck` is clean.

Started by Michiel Westerbeek in [this OS session](https://os.tella.dev/session/os-01a01905-6c95-7000-817d-d6fb618beb9b)


tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `712c4c1`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-102-review)

</details>

tella-butler:
### 🤖 OS review

⚠️ Couldn't prepare the PR checkout to review the diff — it will retry on the next push, or ask OS to review manually.

PR #103, opened 2026-08-19
Fix Windows Runner command execution

happylinks:
A Windows Runner paired today was completely unusable: every `run_on_runner`
call failed at spawn with `ENOENT ... uv_spawn 'powershell.exe'`, and
`opensession runner service install` printed nothing at all and exited 1.
`C:\Windows\System32` was not resolvable on that machine's PATH. The product
defect is real regardless of that one box, and this fixes it.

## Stop trusting PATH for System32 binaries

`powershell.exe` and `schtasks` were resolved through PATH in five places,
which makes the Runner's core function depend on a variable any installer,
group policy or truncated `setx` can damage. New exported helpers
(`windowsSystem32`, `windowsPowerShellPath`, `resolveWindowsShell`,
`resolveWindowsSchtasks`, `runnerExecCommand`) build the path from
`%SystemRoot%` instead, with a fallback chain that ends at today's bare name
so a healthy machine is not regressed. Used by the exec path, the connect-time
disk probe, the schtasks invocation, the Portal spawn, the Runner terminal,
and the scheduled task's `<Command>`.

The task XML is the worst of the five: a bare name there means Task Scheduler
silently launches nothing when PATH is damaged at sign-in, so the Runner just
never appears online with no error anywhere.

Resolving our own spawn only fixes the spawn. The command the shell then runs
is written by a person or an agent and says `git`, `where`, `robocopy`,
`Get-CimInstance`, all of which would still fail inside a shell that started
fine. So `windowsRunnerEnvironment` now also puts `System32`,
`System32\WindowsPowerShell\v1.0` and `System32\Wbem` back on the PATH it
hands to children. The machine's own PATH is left untouched. This is what
makes one bootstrap command enough, with no registry edit and no elevation.

## Stop failing silently

`installRunnerService()` was a try/catch over three platform branches, each
guarded by a capability check. When a guard was false, no branch ran, nothing
threw, nothing was caught, and it returned `false` having printed literally
nothing. Reproduced before the fix, with PATH stripped so the systemctl guard
is false:

```
OLD returned: false | lines printed: 0 | []
NEW returned: false | lines printed: 1
  > fail    Runner service was not installed  systemctl is not on this machine, ...
```

Each guard now names the platform and the missing service manager, win32
points at the likely cause, and `scripts/cli.ts` adds what to do meanwhile.

## CI that would have caught it

`runner-windows` covered install.ps1, the unit tests and the service install,
but never executed a command over the Runner channel, and a GitHub image has a
pristine PATH, so all of the above passed green. The new step spawns through
the same argv the exec path builds with a PATH that cannot reach System32, and
asserts both that the shell starts and that a System32 tool resolves *inside*
the delegated command. The service step now also asserts the registered task
action names a fully qualified PowerShell.

Nine new unit tests cover the resolvers, the PATH repair and the emitted XML
as pure functions, because no Windows box runs that suite locally.

## Docs

`docs/runners.md` gains an "Operating a Windows Runner" subsection: the
`OpenSessionRunner` task name and the `schtasks /Query|/Run|/Delete` commands,
the log at `%USERPROFILE%\.opensession\runner.log`, that
`opensession runner service install` is the fallback and needs no elevation or
PATH editing, that the trigger is a LogonTrigger with `InteractiveToken` (so a
headless box needs autologon to survive a reboot), that
`MultipleInstancesPolicy: IgnoreNew` silently skips the launch while a
foreground `runner run` is going, keeping the machine awake (`powercfg`,
Modern Standby and `PlatformAoAcOverride`, and the NIC power setting that
drops the tailnet while the box stays up), and remote access scoped to
`100.64.0.0/10`. VNC shares the console session an autologon box depends on;
RDP disconnects it.

## Not verified

No live Windows box. The pure functions are unit tested on Linux and the
end-to-end behaviour is asserted by the new CI step on `windows-latest`;
nothing here has been run against a real damaged-PATH machine.

Started by Michiel Westerbeek in [this OS session](https://os.tella.dev/session/os-01a01958-7bbc-7000-bb57-a6ee7dce0250)


tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `a0af2b8`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-103-review)

</details>

tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `a0af2b8`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-103-review)

</details>

tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `a0af2b8`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-103-review)

</details>

tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `a0af2b8`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-103-review)

</details>

tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `9f5a0a2`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-103-review)

</details>

tella-butler:
Safe to merge. The PR consistently resolves Windows system binaries independently of PATH, repairs the delegated child environment, applies the shared command builder across Runner execution surfaces, and adds focused Windows CI coverage. I found no merge-blocking correctness or safety issues.
<sub>Reviewed `4886d4f` · Pi · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-103-review)</sub>

happylinks:
One note on the new CI step, so the next reader does not over-trust it.

The line that reads like a control —

```
bare powershell.exe with System32 off PATH: 0
```

— spawns through `windowsRunnerEnvironment(broken)`, i.e. the **repaired** env, so it is not controlling anything. It will print `0` whether or not the fix is present, and a reader scanning the log could take it as "bare lookup works even when PATH is broken", which is the opposite of the bug.

What the step genuinely proves, and this is enough:

- `resolveWindowsShell()` / `resolveWindowsSchtasks()` return fully-qualified `C:\\Windows\\System32\\...` paths rather than bare names;
- `windowsRunnerEnvironment` puts the three System32 directories back on a `PATH` of `C:\\nope`;
- the real `runnerExecCommand` path round-trips (`exit 0`, `ok-from-runner`) through that environment;
- a System32 tool (`where.exe cmd.exe`) resolves *inside* the delegated command, which is what matters for the commands people actually send (`powercfg`, `netsh`, `git`).

If the control is worth keeping, it should spawn through the **raw** `broken` env, not the repaired one — then it fails today and would have failed before the fix, which is the assertion it looks like it is making.

Second, smaller thing: the Windows job is also green here partly because 9f5a0a2d5 skips three `service.test.ts` assertions on win32. That skip is correct on its own merits (they describe systemd/launchd), but it is noise removal, not evidence for this fix — worth not conflating the two when reading the job status.

PR #105, opened 2026-08-19
Document Runner elevation and the two-instance failure

happylinks:
Follow-up to #103, from actually operating the Windows Runner "Bill" end to end.

Two things cost real time and were documented nowhere:

**Delegated commands run with a UAC-filtered token.** `whoami /groups` reports `BUILTIN\Administrators` as *"Group used for deny only"* even when the operator is an administrator, and the `IsInRole` check returns `False`. The boundary does not fall where the command names suggest — `powercfg /change standby-timeout-ac 0` works fine, `powercfg /hibernate off` does not; installing the scheduled task works, `New-NetFirewallRule` does not; `Get-NetAdapterPowerManagement` fails even to *read*. So the whole "keep the machine awake / install a VNC server / scope a firewall rule to the tailnet" section of the existing docs is silently undoable from a normal Runner. The new section gives the elevated-Runner recipe and says to hand the channel back afterwards rather than leaving a Runner permanently elevated.

**Two live Runners sharing one identity is a distinctive failure.** They take turns owning the control connection and knock each other off. The Runner reads `online` with a `lastSeenAt` that keeps advancing while commands fail with `[Runner disconnected]` — which reads as a network fault, not a duplicate process. `MultipleInstancesPolicy: IgnoreNew` does not prevent it: it only stops the *task* starting a second copy of itself, not a foreground `opensession runner run` alongside a running task.

Docs only; no code change.

Started by Michiel Westerbeek in [this OS session](https://os.tella.dev/session/os-01a01944-f9a3-7000-9e90-6734617d6afd)

### DIFF
--- a/docs/runners.md
+++ b/docs/runners.md
@@ -57,8 +57,9 @@ Read that first when a Runner shows offline.
 
 If `connect` cannot install the task, it says why and
 `opensession runner service install` runs the same installation on its own.
-There is nothing to configure first on Windows, and nothing to elevate: the
-task is per-user and installs from an ordinary PowerShell window.
+There is nothing to configure first on Windows, and the task itself needs no
+administrator rights: it is per-user and installs from an ordinary PowerShell
+window. Machine-wide setup does need elevation, which is its own subject below.
 
 A damaged PATH does not need repairing by hand. The Runner resolves PowerShell
 and `schtasks.exe` at their known location under `%SystemRoot%\System32`, and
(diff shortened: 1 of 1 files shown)


tella-butler:
Not safe to merge yet. The new operational guidance correctly describes elevation and duplicate Runner symptoms, but it also recommends terminating every Bun process on the machine and leaves contradictory guidance about whether a foreground Runner prevents the scheduled task from starting.
_2 inline comments below._
<sub>Reviewed `8f251b2` · Pi · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-105-review)</sub>

docs/runners.md
```diff
@@ -76,6 +77,57 @@ Two behaviours of the task decide how an always-on box has to be set up:
...
+| | installing a service, such as a VNC server |
+
+So the awake settings, the firewall rules and a VNC server all have to be
+driven from a Runner that was started elevated. Start one from an
+**Administrator** PowerShell window, stopping the task first so only one
+instance holds the channel:
+
+```powershell
+schtasks /End /TN OpenSessionRunner
+Get-CimInstance Win32_Process -Filter "Name='bun.exe'" |
```
tella-butler, line 107: 🔴 **P1** — Do not terminate every Bun process on the machine

This command selects processes solely by `Name='bun.exe'`, so it kills unrelated Bun services, development servers, or scripts along with the Runner. For example, a Windows build host running both Open Session and a Bun-based CI worker would lose the CI worker, potentially interrupting writes or corrupting build output. The diagnostic at line 128 has the same problem and can falsely report duplicate Runners when another Bun application is running. Filter `Win32_Process.CommandLine` for this installation's `scripts\cli.ts runner run` invocation before counting or stopping processes.

docs/runners.md
```diff
@@ -76,6 +77,57 @@ Two behaviours of the task decide how an always-on box has to be set up:
...
+  ForEach-Object { Stop-Process -Id $_.ProcessId -Force }
+opensession runner run
+```
+
+When the machine-wide work is finished, close that window and
+`schtasks /Run /TN OpenSessionRunner` to hand the channel back to the ordinary
+per-user task. A Runner left permanently elevated gives every delegated command
+administrator rights it does not need.
+
+**One instance, always.** `MultipleInstancesPolicy: IgnoreNew` only stops the
```
tella-butler, line 117: 🔴 **P1** — Resolve the contradictory MultipleInstancesPolicy guidance

This correctly says `IgnoreNew` does nothing about a foreground process, but lines 76-78 still say the task launch is ignored when `opensession runner run` is active. Task Scheduler cannot identify an unrelated foreground invocation; the server instead replaces the existing connection when another process connects (`runnerWsOpen` closes the previous socket). An operator following the earlier paragraph can therefore start the task beside the foreground Runner and recreate the exact connection flapping this section is intended to prevent. Update the earlier bullet to state that the policy only suppresses duplicate scheduled-task instances and that foreground processes must be stopped separately.

tella-butler (commented):
OS review · `8f251b2`

PR #106, opened 2026-08-19
Give automations a read of the audit digest

tella-butler:
## Why

Every automation runs on the Pi engine since aeb73d59f. Pi ask mode's local tool list is `["read","grep","find","ls"]` (`pi-runner.ts:1907`), those tools are sandboxed to the session workspace, and `automationRunInProcessMcp` does not mount `opensession-web` (which refuses loopback anyway, `web-fetch.ts`). Net effect: the nightly Dreaming reflection, whose entire job is to read yesterday's audit log, has no path to `~/.opensession-audit/*.jsonl` at all. It ran blind.

I verified each of those claims against the code before writing anything; they all hold at `faae15e2e`.

## What

`packages/core/opensession-server/src/server/audit-mcp.ts`, in the shape of the existing `health-mcp.ts` precedent: one in-process MCP server, `opensession-audit`, with one tool, `read_audit_digest`. It takes one optional `date` and returns `buildAuditDigest(date)`, the same roll-up `GET /api/audit/digest` serves. Mounted in `automationRunInProcessMcp` next to `opensession-health`. Automations only, nothing added to interactive runs or any other server set.

The bar it is held to, same as its neighbours: read-only, no writes, no shell, no network, no new env, and the only argument is a date.

- `date` is validated against `/^\d{4}-\d{2}-\d{2}$/` before it becomes a filename component, and a rejected value never reaches the log. Not a path, not a glob, not a range. A blank date reads as an omitted one (yesterday, UTC).
- A missing log returns `ok: false` with the days that do exist, rather than throwing.
- `papercuts` and `sessions` are capped, with a `truncated` block saying what was dropped. If the capped document is still outsized, whole detail sections drop until it fits, and the totals always survive. A busy day cannot return megabytes.

`audit-mcp.test.ts` covers the valid date, the default date, eight malformed dates (including `../../etc/passwd` and a range), the missing-log case, and both truncation paths.

## Checks

- `bun run typecheck`: clean, 0 errors.
- `bun test packages/core/opensession-server/src/server/audit-mcp.test.ts`: 7 pass, 0 fail.
- `bun scripts/check-module-side-effects.ts`: ok, 345 modules import cleanly.

This is a runner/automation-layer change, so it needs a deliberate `systemctl restart opensession` after merge.

**No reviewer was requested automatically. A human needs to be asked directly.**

Started by Dreaming (nightly reflection) in [this OS session](https://os.tella.dev/session/os-01a019ed-3395-7000-b641-92a222bca6d5)

### DIFF
--- a/packages/core/opensession-server/src/server/automations.ts
+++ b/packages/core/opensession-server/src/server/automations.ts
@@ -42,6 +42,7 @@ import { createPapercutsMcpServer } from "../agents/slack/papercuts-tools";
 import { createReportMcpServer } from "../agents/slack/report-tools";
 import { createWorkflowsMcpServer } from "../agents/slack/workflow-tools";
 import { createTurnMcpServer } from "../agents/slack/turn-tools";
+import { createAuditMcpServer } from "./audit-mcp";
 import { createHealthMcpServer } from "./health-mcp";
 import { papercutsEnabledForRepo } from "./papercuts";
 import {
@@ -942,6 +943,14 @@ function automationRunInProcessMcp(
     // refuses loopback, and an unattended ask run has no shell on either
     // engine (src/server/health-mcp.ts).
     "opensession-health": createHealthMcpServer(),
+    // And once more: one read of a single day's rolled-up audit log, whose
+    // only argument is a date validated as YYYY-MM-DD before it becomes a
+    // filename component. Nothing to steer, nothing to escalate with. Here
+    // for the same reason as its neighbour, and it is the nightly reflection
+    // that cannot read the log any other way: web-fetch.ts refuses loopback,
+    // and Pi's ask tools are sandboxed to the session workspace, so
+    // ~/.opensession-audit is unreachable (src/server/audit-mcp.ts).
+    "opensession-audit": createAuditMcpServer(),
   };
 }
 
--- a/packages/core/opensession-server/src/server/audit-mcp.test.ts
+++ b/packages/core/opensession-server/src/server/audit-mcp.test.ts
@@ -1,0 +1,108 @@
+import { describe, expect, it } from "bun:test";
+import { type AuditDigestDeps, auditDigestPayload } from "./audit-mcp";
+
+/** A digest stub with the fields the payload caps or reports on. `build`
+ *  records what it was asked for, so a rejected date can be shown never to
+ *  reach the filename. */
+function deps(
+  digest: Record<string, unknown> | null,
+  asked: string[] = []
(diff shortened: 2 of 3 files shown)


tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `3d92cee`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-106-review)

</details>

tella-butler:
Safe once the P1 below is fixed. The new MCP server provides the intended audit digest, but mounting it on every automation crosses the codebase's established boundary for untrusted automation runs and exposes instance-wide session content.
_1 inline comment below._
<sub>Reviewed `3d92cee` · Pi · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-106-review)</sub>

packages/core/opensession-server/src/server/automations.ts
```diff
@@ -942,6 +943,14 @@ function automationRunInProcessMcp(
...
     // engine (src/server/health-mcp.ts).
     "opensession-health": createHealthMcpServer(),
+    // And once more: one read of a single day's rolled-up audit log, whose
+    // only argument is a date validated as YYYY-MM-DD before it becomes a
+    // filename component. Nothing to steer, nothing to escalate with. Here
+    // for the same reason as its neighbour, and it is the nightly reflection
+    // that cannot read the log any other way: web-fetch.ts refuses loopback,
+    // and Pi's ask tools are sandboxed to the session workspace, so
+    // ~/.opensession-audit is unreachable (src/server/audit-mcp.ts).
+    "opensession-audit": createAuditMcpServer(),
```
tella-butler, line 953: 🔴 **P1** — Do not expose instance-wide session content to every automation

`automationRunInProcessMcp` is used by all automations, including webhook, ticket, and Slack-channel runs whose event payload is untrusted. The digest is not merely aggregate telemetry: `buildAuditDigest` includes the first 200 characters of session prompts, papercut messages verbatim, raw error samples, tool-error samples, and session IDs. For example, a prompt-injected watched-channel message can instruct its automation to read yesterday's digest and place unrelated customer ticket text from `sessions[].firstPrompt` into its report or configured output. This violates the documented automation invariant immediately above this function that in-process tools expose "nothing sensitive readable." Mount this server only for the human-enabled introspection automation, such as behind the existing `selfImprove` gate used by Dreaming, or add a dedicated human-set audit-access capability rather than granting it to every automation.

tella-butler (commented):
OS review · `3d92cee`

PR #107, opened 2026-08-19
Spin off: hand the link to the composer

happylinks:
"Spin off" had three rows and all of them write the prompt for you: Build this, Capture learnings, Analyze session. There was no way to say "start something new, but from here" in your own words.

This adds a fourth row that spins nothing off itself — it opens the new-session composer with a reference to this session already in the draft, caret parked after it. No modal, no branch field, no generated prompt: the composer *is* the form.

The reference written into the draft is the bare session id rather than the URL, because that is the form the composer settles on anyway — a pasted session link is rewritten to its id the moment it lands (`insertPastedSessionId`), and the id is what paints as a chip carrying the session's title. Handing the composer a URL it would immediately shorten would be taking the long way round on purpose. A session whose id predates the minted shape has no bare form that chips, so it keeps its whole URL, which the renderer reads too.

`composerSessionRef` in `lib/share-link.ts` is the new helper (covered by tests); `isMintedSessionId` is now exported from `lib/session-url.ts` so the write side asks the same question the read side already did.

Created by [this OS session](https://os.tella.dev/session/os-01a01ae5-ce77-7000-8874-98ded6c9208f)

### DIFF
--- a/packages/core/opensession-server/src/frontend/components/SpinOffMenu.tsx
+++ b/packages/core/opensession-server/src/frontend/components/SpinOffMenu.tsx
@@ -8,6 +8,8 @@ import type { UnifiedSession, TranscriptEntry } from "../lib/types";
 import { getCurrentUser } from "./UserPicker";
 import { Field, fieldClasses } from "../ui/input";
 import { noAutofill } from "../lib/composer-autofill";
+import { composerSessionRef } from "../lib/share-link";
+import type { NewSessionPrefill } from "../lib/new-session-link";
 
 type Flavor = "build" | "learnings" | "analyze";
 
@@ -16,6 +18,9 @@ interface Props {
   entries: TranscriptEntry[];
   send: (msg: any) => void;
   connected: boolean;
+  /** Open the new-session composer prefilled — the one flavor that writes no
+      prompt of its own. */
+  onOpenNewSession: (prefill: NewSessionPrefill) => void;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


tella-butler:
Safe once the mode-locking issue below is fixed. The PR correctly adds a session reference to the new-session draft, but referencing an Ask session prevents the user from choosing Code, undermining the composer-as-form workflow.
_1 inline comment below._
<sub>Reviewed `d7c82dd` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-107-review)</sub>

packages/core/opensession-server/src/frontend/components/SpinOffMenu.tsx
```diff
@@ -48,6 +63,17 @@ export function SpinOffMenu({ session, entries, send, connected }: Props) {
     }
   }
 
+  function openLinkedSession() {
+    onOpenNewSession({
+      // A reference, not a transcript: the new session reads this one through
+      // the link when it needs to, instead of carrying a copy that goes stale.
+      // The trailing space parks the caret clear of the chip.
+      prompt: `${composerSessionRef(session)} `,
+      mode: session.mode ?? "code",
```
tella-butler, line 72: 🟠 **P2** — Prefilling the mode also locks the composer to that mode

`openPrefilledSession` passes this value to `NewSession.forceMode`. `NewSession` hides its Ask/Code toggle whenever `forceMode` is present, and specifically fixes `permission` to Ask when `forceMode === "ask"` (`NewSession.tsx:426-429,1463`). Therefore, selecting **Reference this session** from an Ask session opens a composer where Code cannot be selected. A user who wants to write their own coding prompt with only this session reference cannot use the new workflow; **Build this** is not equivalent because it copies and generates context. Add a distinct initial/default mode prefill that leaves the toggle available, and use that here instead of `forceMode`.

tella-butler (commented):
OS review · `d7c82dd`

PR #108, opened 2026-08-20
Fix outward-effect tool identifiers

tella-butler:
## Summary
- replace stale Slack reach-tool identifiers with the mounted post and thread-reply tools
- recognize Linear's catalog-backed `save_comment` and `save_issue` writes
- update one-off Slack reminder instructions to call `slack_reply_to_thread` with its current arguments
- pin the exact Pi event spellings while keeping reads and unrelated `save_*` tools excluded

## Evidence
- `scripts/mcp-slack.ts` and the mounted tool catalog expose `slack_post_message` and `slack_reply_to_thread`
- Pi mounts MCP tools as `<server>_<tool>`, producing `slack_slack_post_message` and `slack_slack_reply_to_thread`
- Linear's mounted catalog describes `save_comment` and `save_issue` as create-or-update writes

## Tests
- `bun test packages/core/opensession-server/src/server/turn-outcome.test.ts`
- `bunx tsc --noEmit --pretty false` (blocked by a pre-existing `Setup.tsx` `compact` prop error outside this change)

Started by Dreaming (nightly reflection) in [this OS session](https://os.tella.dev/session/os-01a01d7c-40ae-7000-a380-524b468ca0a9)

### DIFF
--- a/packages/core/opensession-server/src/server/turn-outcome.ts
+++ b/packages/core/opensession-server/src/server/turn-outcome.ts
@@ -70,10 +70,11 @@ const REACH_TOOLS = [
   "mcp__plain__create_note",
   "mcp__plain__reply_to_thread",
   // Slack — a message in a channel or DM.
-  "mcp__slack__conversations_add_message",
-  // Linear — a comment or a new issue on the team's board.
-  "mcp__linear__create_comment",
-  "mcp__linear__create_issue",
+  "mcp__slack__slack_post_message",
+  "mcp__slack__slack_reply_to_thread",
+  // Linear — a comment or an issue created or updated on the team's board.
+  "mcp__linear__save_comment",
+  "mcp__linear__save_issue",
   // Gmail — outbound mail (a draft is not delivery; sending is).
   "mcp__gmail__send_email",
   // Our own in-process servers: a published report, and asking a teammate.
--- a/packages/core/opensession-server/src/agents/slack/admin-tools.ts
+++ b/packages/core/opensession-server/src/agents/slack/admin-tools.ts
@@ -443,9 +443,9 @@ export function createAdminMcpServer(ctx: AdminToolContext) {
           let mcpServers = args.mcpServers;
           if (replyInThread) {
             prompt +=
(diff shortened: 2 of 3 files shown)


tella-butler:
Safe once the P2 below is fixed. The catalog identifiers are correct, but Pi now emits the outer `mcp_call` dispatcher as the tool event, so the new Slack and Linear entries still never reach this classifier during real runs.
_1 inline comment below._
<sub>Reviewed `bfed06e` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-108-review)</sub>

packages/core/opensession-server/src/server/turn-outcome.ts
```diff
@@ -70,10 +70,11 @@ const REACH_TOOLS = [
   "mcp__plain__create_note",
   "mcp__plain__reply_to_thread",
   // Slack — a message in a channel or DM.
-  "mcp__slack__conversations_add_message",
-  // Linear — a comment or a new issue on the team's board.
-  "mcp__linear__create_comment",
-  "mcp__linear__create_issue",
+  "mcp__slack__slack_post_message",
```
tella-butler, line 73: 🟠 **P2** — Unwrap the compact MCP dispatcher before classifying effects

These identifiers are not the names observed by the ledger in current Pi runs. `pi-runner.ts` registers only `mcpBridge.discoveryTools`, which `pi-mcp-bridge.ts` defines as `mcp_search` and `mcp_call`; `mcp_call` invokes the selected bridged tool directly inside its executor, so it does not generate a nested `tool_execution_start`. For example, posting through `mcp_call({ name: "slack_slack_post_message", ... })` emits a `tool_use` whose `toolName` is `mcp_call` and whose input contains the real name. `agent-runner.ts` therefore passes `mcp_call` to `isReachTool`, gets `false`, and records a false `silent-drop` despite the Slack message being sent. The same issue affects the new Linear entries. In the observer, extract `event.toolInput.name` when `event.toolName === "mcp_call"`, then classify and record that identifier; add an integration-shaped test using this actual event form.

tella-butler (commented):
OS review · `bfed06e`

PR #109, opened 2026-08-20
Fix Pi audit digest turn aggregation

tella-butler:
## Summary
- treat `session_turn_metric` as the logical Pi turn terminal and aggregate the session-bearing `pi_turn` attempts that precede it
- exclude one-shots, continuation-only records, and non-terminal account-switch records from session turn totals
- prefer the Pi logical terminal over mirrored generic result/error events during mixed-format periods
- enrich session run kind and mode when runner metadata first appears instead of freezing early unknown values

## Evidence
- `bun test packages/core/opensession-server/src/server/audit.test.ts packages/core/opensession-server/src/server/audit-mcp.test.ts`
- `bunx tsc --noEmit --pretty false` (the only reported error is the pre-existing `Setup.tsx:343` compact prop mismatch outside this PR)
- replaying `audit-2026-08-19.jsonl` changes unknown run-kind sessions from 499/501 to 14/501 and reports 641 logical turns, 68 terminal errors, and $2,399.48 after including Pi turns

Started by Dreaming (nightly reflection) in [this OS session](https://os.tella.dev/session/os-01a01d7c-1c84-7000-808d-0149dff0d00e)

### DIFF
--- a/packages/core/opensession-server/src/server/audit.ts
+++ b/packages/core/opensession-server/src/server/audit.ts
@@ -225,7 +225,14 @@ export function buildAuditDigest(date: string): Record<string, unknown> | null {
   ensureAuditDir();
   const path = `${AUDIT_DIR}/audit-${date}.jsonl`;
   if (!/^\d{4}-\d{2}-\d{2}$/.test(date) || !existsSync(path)) return null;
+  return buildAuditDigestFromLines(date, readFileSync(path, "utf-8"));
+}
 
+/** Pure digest builder used by focused mixed-format fixtures. */
+export function buildAuditDigestFromLines(
+  date: string,
+  contents: string,
+): Record<string, unknown> {
   interface SessionAgg {
     id: string;
     runKind: string;
@@ -241,15 +248,22 @@ export function buildAuditDigest(date: string): Record<string, unknown> | null {
     costUsd: number;
   }
   const sessions = new Map<string, SessionAgg>();
+  const eventSessionId = (e: Record<string, unknown>): string =>
+    String(
+      e.session_id ||
+        e.bks_session_id ||
+        (e.msg === "pi_turn" ? e.session : "") ||
+        "",
+    );
   const sessionOf = (e: Record<string, unknown>): SessionAgg | null => {
-    const id = String(e.session_id || e.bks_session_id || "");
+    const id = eventSessionId(e);
     if (!id) return null;
     let s = sessions.get(id);
     if (!s) {
       s = {
         id,
-        runKind: String(e.run_kind || "?"),
-        mode: String(e.mode || "?"),
+        runKind: "?",
+        mode: "?",
         models: new Set(),
         firstPrompt: "",
         turns: 0,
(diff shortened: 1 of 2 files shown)


tella-butler:
Safe once the two P1 aggregation gaps below are fixed. The new logical-turn boundary is sound, but mixed old-engine-to-Pi fallbacks still double-count errors, and failed Pi attempts lose their incurred cost.
_2 inline comments below._
<sub>Reviewed `b818b01` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-109-review)</sub>

packages/core/opensession-server/src/server/audit.ts
```diff
@@ -302,14 +320,56 @@ export function buildAuditDigest(date: string): Record<string, unknown> | null {
...
+  // This excludes one-shots and lets mixed-format deployments prefer Pi's
+  // logical terminal without also counting their mirrored generic terminal.
+  for (const e of parsedEvents) {
+    const id = eventSessionId(e);
+    if (!id) continue;
+    if (e.msg === "pi_turn" && e.direction === "out" && !e.kind) {
+      const attempts = pendingPi.get(id) || [];
+      attempts.push(e);
+      pendingPi.set(id, attempts);
+    }
```
tella-butler, line 354: 🔴 **P1** — Generic terminals before the Pi fallback are still counted

Generic terminals are collected only after `pendingPi` exists. The retired runner calls `turnEvent({ kind: "error" })` before yielding its error, and `runAgentInner` starts the fallback model only after consuming that event. Therefore an OpenCode attempt that fails before a Pi fallback produces `generic error -> pi_turn -> session_turn_metric`; this code leaves the generic error outside `genericTerminalsToSkip`, then also counts the metric as the logical terminal. A successful fallback consequently reports one turn and one error instead of one successful turn. Collect generic terminals for the whole interval since the previous `session_turn_metric`, then discard that interval's generic terminals whenever the metric has Pi attempts.

packages/core/opensession-server/src/server/audit.ts
```diff
@@ -321,6 +381,45 @@ export function buildAuditDigest(date: string): Record<string, unknown> | null {
       continue;
     }
     const s = sessionOf(e);
+    const piTurn = piTurns.get(e);
+    if (piTurn) {
+      turns++;
+      const failed = e.outcome === "failed";
+      const cost = piTurn.attempts.reduce(
+        (sum, attempt) => sum + (Number(attempt.total_cost_usd) || 0),
+        0,
```
tella-butler, line 390: 🔴 **P1** — Failed Pi attempts do not carry the cost being summed

The fixture gives failed attempts `total_cost_usd`, but the real emitter does not. In `pi-runner.ts`, `endTurn` includes token and cost fields only for `terminal.type === "done"`; error terminals emit only `{ error }`. For example, if a model consumes $1 before a transient failure and its fallback succeeds for $0.50, this aggregation reports $0.50 rather than $1.50. Include the accumulated usage and `total_cost_usd` on failed `pi_turn` terminal records before relying on this sum, and add a fixture matching the actual emitted failure shape.

tella-butler (commented):
OS review · `b818b01`

PR #116, opened 2026-08-21
Read the tool a run called, not the dispatcher it called it through

tella-butler:
Every unattended turn verdict has been `silent-drop` since 2026-08-19 09:01 UTC. On 08-20 that was 214 rows out of 214, not one carrying an `effects` field, against 8 successful `finish_silently` calls the same day. Both halves of the ledger were reporting falsely, for two different reasons that happen to share one root: **the ledger was reading the dispatcher envelope instead of the call inside it.**

## Effects

Pi does not hand the model one tool per bridged MCP tool. Since the catalog was compacted (`d6814a422`), `createPiMcpBridge` keeps the catalog server-side and exposes exactly two dispatcher tools, `mcp_search` and `mcp_call` (`pi-mcp-bridge.ts` — `const discoveryTools = tools.length ? [searchCatalog, callCatalog] : []`; `pi-runner.ts` puts only `mcpBridge.discoveryTools` into `customTools`).

So pi's `tool_execution_start` never carries `slack_slack_post_message`. It carries `toolName: "mcp_call"` with the real id inside `args.name`, `pi-runner.ts:2307` copies that verbatim into the `tool_use` StreamEvent, and `agent-runner.ts` asked `isReachTool("mcp_call")`, which is correctly `false`.

Measured from the transcript store, on the session that demonstrably posted to Slack and was recorded as reaching nobody (`os-01a01d6e-ee3f-7000-beab-99fa7c905326`, seq 267):

```json
{"type":"tool_use","toolName":"mcp_call",
 "toolInput":{"name":"slack_slack_post_message",
              "arguments":{"channel_id":"C0AFQ7PV057","text":"*🌙 Dreaming — 2026-08-20*..."}}}
```

That session's 135 `tool_use` rows contain only `ls, mcp_search, mcp_call, bash, read, grep` — no bridged tool name at all. A pre-cutover `reached` session has a literal `opensession-report_publish_report` row.

**Why 2026-08-19 09:01 specifically:** `aeb73d59f` "Move automations and one-shots to Pi" (09:41) and its restart. The bridge compaction had been live since 08-14 and was harmless until unattended runs reached it. The 09:01 → 10:00 gap in the hourly verdict histogram is that restart.

## Declarations

`createTurnMcpServer({ turnKey: sessionId })` is built in the **server** process (`automations.ts`, and `interactive-mcp.ts`'s `automationSessionMcp` for the hosted/resume builder). Since the same commit, `automations.ts` dispatches through `runAgentHosted`, which passes only `proxyMcpServers` — names, not instances. The host turns those into stdio `mcp-proxy` children (`runner-host/host.ts` `proxyMcpConfigs`) that call back over run-rpc, so the tool body executes in the server. `beginTurn`/`endTurn` run in the **host**. `recordDeclaration` therefore looked up `ledgers` in the server process, found nothing, and `if (!ledger || ledger.closed) return;` swallowed it.

## The fix

Both halves are the same read, so one observation point repairs both — the declaration is *already* in the host's own event stream as `mcp_call{name:"opensession-turn_finish_silently", arguments:{reason}}` (measured, seq 10 of `os-01a01ed4-55a2-7001-b1b8-9861a6070cb0`).

`observedToolCall` unwraps `mcp_call` to the tool underneath; `observeToolCall` folds that into the ledger as an effect or a declared silence. `agent-runner.ts` calls it at the one place it already watched.

Chosen over relaying records to the server's ledger over a new run-rpc frame because it needs **no wire change at all**: ~100 lines in one module versus a new protocol frame, a `HostRunControl` method, a host handler, a registry lookup, and a delivery race against turn end.

## Deliberately not changed

- `REACH_TOOLS` and `isReachTool` are untouched and stay exactly as explicit as before. They are simply asked about the tool the run called rather than the dispatcher it called it through.
- `recordDeclaration` still refuses to create a ledger it was not given.
- No verdict is suppressed, nothing defaults to `reached`, `automation` stays in `CHECKED_KINDS`.
- The tool-side `recordDeclaration` stays — correct for in-process and fallback runs, and last-wins makes the duplicate a no-op.
- Nothing here needs a restart to be *safe*. The change is inert until new hosts spawn.

## Tests

`turn-outcome.test.ts` gains 22 tests covering the cross-process, dispatcher-wrapped case in the exact event shapes a hosted pi automation emits: an effect through `mcp_call`, `finish_silently`/`stay_silent` through `mcp_call` (with and without a reason), the harmless in-process duplicate, effects still outranking declarations, and the negative cases — reads and papercuts through the dispatcher are still not effects, and a malformed dispatcher call cannot manufacture one. 33/33 pass.

## Adjacent, not fixed here

`createSlackPostScanner` (`slack-links.ts`) has the identical blindness: it regexes `toolName` for `slack_*post_message`, which is now `mcp_call`, so automation Slack-thread linking broke on the same date from the same cause. Left out to keep this diff narrow; worth its own change.

Started by Dreaming (nightly reflection) in [this OS session](https://os.tella.dev/session/os-01a022c6-5595-739d-8a8b-6930cf195543)

### DIFF
--- a/packages/core/opensession-server/src/server/turn-outcome.ts
+++ b/packages/core/opensession-server/src/server/turn-outcome.ts
@@ -93,6 +93,106 @@ export function isReachTool(toolName: string | undefined): boolean {
   return !!toolName && REACH_TOOL_IDS.has(toolName.toLowerCase());
 }
 
+/** The `opensession-turn` server's two tools, in the spellings a run reports. */
+const SILENCE_TOOL_IDS: ReadonlyMap<string, SilenceTool> = new Map(
+  (["finish_silently", "stay_silent"] as SilenceTool[]).flatMap(
+    (tool) =>
+      [
+        [`mcp__opensession-turn__${tool}`, tool],
+        [`opensession-turn_${tool}`, tool],
+      ] as Array<[string, SilenceTool]>
+  )
+);
+
+export function silenceToolFor(toolName: string | undefined): SilenceTool | undefined {
+  return toolName ? SILENCE_TOOL_IDS.get(toolName.toLowerCase()) : undefined;
+}
+
+export interface ObservedToolCall {
+  name: string;
+  args: Record<string, unknown>;
+}
+
+function asRecord(value: unknown): Record<string, unknown> {
+  return value && typeof value === "object" && !Array.isArray(value)
+    ? (value as Record<string, unknown>)
+    : {};
+}
+
+/**
+ * The tool a `tool_use` event is really about.
+ *
+ * Pi does not hand the model one tool per bridged MCP tool. The catalog is
+ * kept server-side and reached through two dispatcher tools, `mcp_search` and
+ * `mcp_call` (createPiMcpBridge, src/server/pi-mcp-bridge.ts) — so a Slack
+ * post arrives as `mcp_call` with `{name: "slack_...", arguments: {...}}`,
+ * and the tool that was actually called sits one level down.
+ *
+ * The ledger read the envelope. That was harmless while unattended runs used
+ * an engine that named the tool directly, and became total the day automations
+ * moved to Pi (2026-08-19): 214 turn_outcome rows the following day, every one
+ * a silent-drop, not one carrying an effect, against 8 successful
+ * `finish_silently` calls that same day.
+ *
+ * Unwrapping here rather than widening REACH_TOOLS is the point. The reach
+ * list stays exactly as explicit as it was; it is simply asked about the tool
+ * the run called instead of the dispatcher it called it through.
+ */
+export function observedToolCall(event: {
+  toolName?: string;
+  toolInput?: unknown;
+}): ObservedToolCall | undefined {
+  const outer = event.toolName;
+  if (!outer) return undefined;
+  const input = asRecord(event.toolInput);
+  if (outer.toLowerCase() !== "mcp_call") return { name: outer, args: input };
+  // A dispatcher call with no resolvable target names no tool at all — a
+  // malformed call, not something the run reached anybody with.
+  const inner = input.name;
+  if (typeof inner !== "string" || !inner) return undefined;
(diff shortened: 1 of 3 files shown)


tella-butler:
Safe once the P1 below is fixed. The dispatcher unwrapping repairs declarations and supported effect IDs, but the demonstrated Slack posting path still produces `silent-drop`, so a central stated goal remains unmet.
_1 inline comment below._
<sub>Reviewed `a9c1544` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-116-review)</sub>

packages/core/opensession-server/src/server/turn-outcome.ts
```diff
@@ -93,6 +93,106 @@ export function isReachTool(toolName: string | undefined): boolean {
...
+ * kinds and never carry a ledger at all (CHECKED_KINDS above).
+ */
+export function observeToolCall(
+  key: string | undefined,
+  event: { toolName?: string; toolInput?: unknown }
+): void {
+  if (!key) return;
+  const call = observedToolCall(event);
+  if (!call) return;
+  if (isReachTool(call.name)) {
```
tella-butler, line 183: 🔴 **P1** — The actual Slack post tool still is not recognized as an effect

After unwrapping, the measured call becomes `slack_slack_post_message`, but `isReachTool` only recognizes `slack_conversations_add_message`. The repository's Slack MCP confirms this naming: `scripts/mcp-slack.ts` exposes `slack_post_message` and `slack_reply_to_thread`, and `bridgedTool` prefixes the server name to produce `slack_slack_post_message` and `slack_slack_reply_to_thread`. Therefore the PR's concrete successful Slack example still falls through here and ends as `silent-drop`. Add both current Slack tool IDs to `REACH_TOOLS` and cover the measured dispatcher-wrapped name in the tests.

tella-butler (commented):
OS review · `a9c1544`

PR #117, opened 2026-08-21
Allow actor-owned compatibility writes

jfrolich:
## Summary
- let synchronous compatibility writes borrow the SessionKernel actor's existing physical writer while a durable command owns the mailbox
- persist borrowed effects directly instead of opening a second command lease that cannot synchronously wait
- cover detached run-state, transcript-style, and outbox writes during an active command

## Incident
After opensession#101 deployed, detached callbacks and queue drains hit `Session mutation raced the … mailbox` while long-lived prompt and recovery commands held the actor lease. This failed session creation, transcript persistence, queue drains, and recovery settlement across multiple sessions.

## Verification
- `bun run typecheck`
- 112 focused kernel, create, host, route, and side-effect tests
- `git diff --check`

### DIFF
--- a/packages/core/opensession-server/src/server/session-kernel/actor-client.test.ts
+++ b/packages/core/opensession-server/src/server/session-kernel/actor-client.test.ts
@@ -1,9 +1,16 @@
 import { afterEach, describe, expect, test } from "bun:test";
 import { SessionKernelActorClient } from "./actor-client";
+import {
+  __setSessionKernelStoreForTest,
+  installSessionKernelActor,
+  sessionKernel,
+} from "./kernel";
 import { SESSION_KERNEL_MAX_QUEUED_PER_SESSION } from "./actor-protocol";
 
 let client: SessionKernelActorClient | undefined;
 afterEach(() => {
+  installSessionKernelActor(undefined);
+  __setSessionKernelStoreForTest(undefined);
   client?.terminate();
   client = undefined;
 });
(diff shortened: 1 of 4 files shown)


tella-butler:
Safe to merge. The PR correctly lets synchronous compatibility writes reuse the actor-owned writer during an active command, while preserving normal lease settlement and directly persisting borrowed outbox effects. The changed return contract is propagated through the client, and the focused tests cover run-state, transcript-style, and outbox behavior without exposing a verified regression.
<sub>Reviewed `1f11292` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-117-review)</sub>

happylinks:
Confirming this from the live instance: I hit `Session mutation raced the … mailbox` on my own session and traced it to the same place. Measurements from `journalctl` on the box, 30-minute window before this PR landed on the branch:

| count | call site |
|---|---|
| 36 | `transitionRunState` (run-state.ts:257) → `kernel.setRunState` |
| 30 | `settleRecovery` (agent-runner.ts:1156/1163) |
| 21 | `recordRunOutcome` (session-cache.ts:578) |
| 13 | `SessionOwnedMap.set` (owned-map.ts:53) from `drainQueueInner` (run-session.ts:1039) |
| 6 | `openCreatedSession` (session-create.ts:1012) |

~57 throws / 30 min across ~25 distinct sessions, all through `mutateSync` → `beginSync` → `callStore`. First occurrence 09:34 UTC, right after opensession#101 deployed. User-visible as "Session setup failed", "Run failed", and suppressed transcript appends.

I agree with the borrow, and I want to record why the obvious alternative is not available. Making `beginSync` queue instead of throw is a **guaranteed process-wide deadlock**, not a slow path: `Atomics.wait` parks the gateway thread, and the only thing that can release the lease is the active handler's continuation, which runs on that same parked thread. Nothing can ever notify the waiter. Worth putting in the comment above the borrow so nobody "fixes" it later.

The reason the borrow is safe is also worth stating explicitly: both sides run on one JS thread, so a borrowed write can only execute while the active handler is suspended at an `await`. The interleaving granularity is await points plus per-RPC serialization in the Worker, which is exactly the model these compat call sites were written under before #101. This restores a prior guarantee rather than inventing a new one.

Two things I'd change before merge:

1. **Tombstone check on the borrowed path.** The early `return { duplicate: false, borrowed: true }` jumps over the `store.isTombstoned(sessionId)` check that both `grant` and the normal `beginSync` path do. `mutateSync` calls `assertWritable()` on the gateway first, so this is mostly covered, but that check is a separate RPC and leaves a TOCTOU window — a borrowed write can land after a tombstone committed. Replicate the check in the borrowed branch and **no-op** rather than throw (throwing here would recreate this incident for the delete case).

2. **Batch the effect enqueue.** The `for (const effect of borrowed.effects)` loop is N separate Atomics RPCs, so a crash mid-loop persists a prefix. One `enqueueOutboxMany` worker call makes it a single serialized unit. `effectKey` dedup is sound as-is (`ON CONFLICT(session_id, kind, effect_key) DO NOTHING`, and keys are caller-supplied rather than derived from the borrowed `requestId`), so duplicates are not a concern — partial enqueue is.

Non-blocking, for the follow-up rather than this PR: the root cause is that a long-lived command holds the lease for an entire agent turn, so this shape will keep recurring in new costumes. The architectural fix is an admission-only lease (hold it for durable admission and final settlement, not the whole turn) plus CAS-style run-state transitions in the store, after which the borrow path can shrink or disappear. Also worth noting `recordChange` is a separate RPC after the mutation, so there is a crash window where a mutation is visible without its change record — relevant if that feeds cache invalidation or sync cursors.

Routing these sites through async `dispatch` instead is a dead end, for the record: they cannot await, so dispatch means fire-and-forget, which queues the write behind the multi-minute lease. A run-state transition that becomes visible after the turn ends is wrong, not just late.

Reviewed from https://os.tella.dev/session/os-01a023b0-84dd-77c1-a76c-763b4c25bb25 with a second opinion from the Dial oracle.

PR #119, opened 2026-08-21
Remove session mailbox leases and own delivery state in the actor

jfrolich:
> This is the lease-removal and delivery/ask ownership foundation. Create, turn, transcript and session-file effects still use compatibility adapters and will move in follow-up slices.

## Summary
- replace per-session gateway leases with bounded, fenced execution IDs while keeping physical compatibility effects ordered after durable admission
- move prompt queues, dispatches, steer checkpoints and blocking ask facts into typed Worker reducers backed by schema 7 SQLite aggregates
- atomically claim or restore queue batches and checkpoint steers before touching a runner, including fail-closed restart recovery for ambiguous steering
- import legacy queue and ask JSON once, remove the generic runtime maps, and keep process-only resolver, timer and Stop handles explicitly ephemeral
- keep settled-session reads on a synchronized gateway projection instead of adding synchronous actor IPC to the session-cache hot path

## Safety properties
- duplicate in-flight request IDs attach as bounded waiters to one execution
- stale or unknown execution settlement fail-stops the actor
- replay-safe interrupted effects re-admit; ambiguous non-replay-safe effects remain indeterminate
- delivery and ask changes update their aggregate and session change stream transactionally
- queue batch claim/failure cannot leave a prompt in neither the queue nor dispatch
- a crash around runner steering recovers through a pending-steer receipt instead of redelivering the message
- failed queue-owned starts restore their claimed dispatch instead of acknowledging it away
- queued attachments are persisted as durable file references rather than inline actor payloads
- schema rollback remains fenced by `user_version = 7`

## Verification
- 185 focused SessionKernel, queue, ask, run-state, run-host and create-plan tests pass
- 379 server/executor modules import without side effects
- `git diff --check`
- `bun run typecheck`


tella-butler:
Website build fix verification at both shipped widths. The restored phone glyph appears beside **Native iOS app** with the existing icon system’s stroke weight.

**Desktop**
Website feature grid at desktop width

**Phone**
Website feature grid at phone width

tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `43cc10d`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-119-review)

</details>

tella-butler:
<details>

Safe once the two P1 delivery regressions below are fixed. The migration establishes durable actor-owned queue and ask aggregates, but ordinary runner failures can now drop claimed prompts, and valid large image prompts exceed the actor projection transport limit and wedge their queues.
_2 inline comments below._
<sub>Reviewed `43cc10d` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-119-review)</sub>

</details>

packages/core/opensession-server/src/server/run-session.ts
```diff
@@ -1035,11 +1043,7 @@ async function drainQueueInner(sessionId: string): Promise<void> {
 		} catch (e) {
 			// The batch was already spliced out and persisted away — put it back at
 			// the front of the queue so a throw doesn't lose the messages.
-			acknowledgePromptDispatch(sessionId, promptEntryId, false);
-			const current = promptQueues.get(sessionId) || [];
-			promptQueues.set(sessionId, [...batch, ...current]);
-			persistQueues();
-			broadcastQueue(sessionId);
+			failPromptDispatch(sessionId, promptEntryId);
```
tella-butler, line 1047: 🔴 **P1** — Restore the batch after runSessionPrompt acknowledges a failed dispatch

`runSessionPrompt` acknowledges `durablePromptEntryId` in its catch block at line 1750 before rethrowing. Consequently, when a claimed queue turn fails, such as workspace revival or runner startup throwing before a journal is established, this `failPromptDispatch` finds no active dispatch and returns false. The batch has already been removed from `queued`, so the user's prompt is no longer drainable. The previous implementation explicitly reinserted the batch in this catch path. Preserve the dispatch until this caller can fail it atomically, or fall back to restoring `batch` when `failPromptDispatch` returns false while retaining the direct-send failure behavior.

packages/core/opensession-server/src/server/session-kernel/kernel.ts
```diff
@@ -63,6 +74,95 @@ type CommandContext = {
...
+  else
+    result = store.failDeliveryDispatch(
+      request.sessionId,
+      request.promptEntryId,
+    );
+  if (request.op === "snapshot") {
+    projection.set(request.sessionId, result as DurableDeliveryState);
+  } else if ("sessionId" in request) {
+    const snapshot = actor
+      ? actor.decideDelivery({ op: "snapshot", sessionId: request.sessionId })
```
tella-butler, line 150: 🔴 **P1** — Do not return full attachment-bearing queues through the 8 MB sync buffer

Every delivery mutation now immediately fetches the full session snapshot through `decideDelivery`. That synchronous RPC uses the fixed `LARGE_OUTPUT_BYTES = 8 * 1024 * 1024` buffer, while the verified WebSocket path accepts up to 50 MB of image data and stores its base64 data URLs in `QueueItem.images`. A valid 10 MB image therefore commits the queue row, then the snapshot response exceeds the buffer and `enqueuePrompt` throws before arming its drain. Subsequent `promptQueues.get` and restart hydration also cannot read that aggregate, leaving the prompt and potentially startup recovery wedged. Keep large attachment payloads out of synchronous projection responses, return them through a transport sized or chunked for the accepted aggregate, or persist references instead of inline data.

tella-butler (commented):
OS review · `43cc10d`

packages/core/opensession-server/src/server/run-session.ts
jfrolich: Fixed in 990a2e490. `runSessionPrompt` now retires only a direct dispatch it created itself on failure. A queue-owned dispatch remains intact for `drainQueueInner` to restore atomically.

packages/core/opensession-server/src/server/session-kernel/kernel.ts
jfrolich: Fixed in 990a2e490. Queue and steer intake now stage inline images and files to durable upload references before actor persistence. Read-only actor snapshots also negotiate an exactly-sized response buffer, with a 128 MB fail-stop ceiling, so legacy large aggregates can be migrated without wedging recovery.

tella-butler:
<details>

Safe once the P1 below is fixed. The two previously reported dispatch-recovery and snapshot-buffer issues are addressed, but the REST/session-control steer path still persists inline attachment bodies and can exceed the new actor snapshot ceiling.
_1 inline comment below._
<sub>Reviewed `990a2e4` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-119-review)</sub>

</details>

packages/core/opensession-server/src/server/session-control-wiring.ts
```diff
@@ -245,21 +245,42 @@ registerSessionControl({
 				// (and never steer regardless): the in-thread answer mirror only fires
 				// on a turn that carries the slackReplyTo, and a steered message can't
 				// (it folds into a turn that's already running).
+				const steerItem = {
+					id: deliveryId,
+					content,
+					user,
+					images: opts?.imageUrls,
```
tella-butler: 🔴 **P1** — Stage images before persisting session-control steers

`routes/sessions.ts` passes `asDataUrlList(body.images)` into `opts.imageUrls`, so this path can still put full base64 images into the actor through `prepareQueuedSteer`. For example, two REST prompts carrying roughly 49 MB images while a long-running turn accepts both steers leave over 128 MB of base64 data in the delivery aggregate. The projection snapshot performed after the second mutation then exceeds `MAX_DYNAMIC_OUTPUT_BYTES`; `callSync` marks the actor dead, and subsequent kernel operations fail until restart. The WebSocket path now avoids this by calling `durableQueueItem` first. Apply the same staging to `steerItem` here before `prepareQueuedSteer`, while continuing to pass `opts.images` to the live runner.

tella-butler (commented):
OS review · `990a2e4`

packages/core/opensession-server/src/server/session-control-wiring.ts
tella-butler: Fixed in 95d523721. The session-control steer path now stages inline image bodies with `durableQueueItem` before `prepareQueuedSteer` persists actor state, while the live runner still receives the original parsed images. A structural ownership test fences this ordering.

tella-butler:
Safe once the P2 attachment leak below is fixed. The prior blocking issue is addressed: session-control steers now persist durable image references before runner delivery. The new staging placement, however, duplicates inline images on queue-only paths.
_1 inline comment below._
<sub>Reviewed `95d5237` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-119-review)</sub>

tella-butler (commented):
OS review · `95d5237`

packages/core/opensession-server/src/server/session-control-wiring.ts
```diff
@@ -245,21 +245,42 @@ registerSessionControl({
 				// (and never steer regardless): the in-thread answer mirror only fires
 				// on a turn that carries the slackReplyTo, and a steered message can't
 				// (it folds into a turn that's already running).
+				const steerItem = durableQueueItem(id, {
```
tella-butler: 🟠 **P2** — Do not stage steer attachments on queue-only paths

`durableQueueItem` writes inline image data to disk here before checking whether steering is eligible. For a verified REST request containing a data-URL image with `busyMode: "queue"`, this stages the image once, the steer condition rejects the queue-only request, and `enqueuePrompt` stages the original `opts.imageUrls` again. Only the second path is persisted in actor state, leaving the first file permanently untracked. A single request can therefore consume twice the accepted image size, up to roughly 100 MB for the 50 MB limit. Move `durableQueueItem` inside the branch guarded by the static steer requirements (`busy`, `slackReplyTo`, and `hasFiles`), then fall through to `enqueuePrompt` without pre-staging when steering is ineligible.

tella-butler: Fixed in 238ecb5b2. Inline images are now staged only after the request passes the static steer eligibility checks. Queue-only, Slack-thread, and file-bearing deliveries fall through with the original attachments and are staged once by `enqueuePrompt`; the ownership test now fences eligibility before staging.

PR #122, opened 2026-08-21
Unify file drag ownership

kentdebruin:
## Summary
- remove file drag state and drop handlers from the shared session composer
- keep one app-wide drop target for regular sessions
- let the foreground Desk composer own drops without activating the session underneath

## Verification
- `bun test packages/core/opensession-server/src/frontend/lib/file-drag.test.ts`
- `bun run typecheck`
- verified the app-wide drop overlay at desktop and phone widths

Created by [this OS session](https://os.tella.dev/session/os-01a02495-4afb-7431-9a61-c42db90a73a8)

### DIFF
--- a/packages/core/opensession-server/src/frontend/components/Composer.tsx
+++ b/packages/core/opensession-server/src/frontend/components/Composer.tsx
@@ -54,7 +54,6 @@ import {
   IconStopSquare,
   IconPencil,
   IconTrash,
-  IconArrowUpToLine,
 } from "./icons";
 import {
   composerBox,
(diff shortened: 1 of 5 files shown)


tella-butler:
Safe once the P2 regression below is fixed. The PR correctly transfers app-wide drop ownership between regular sessions and the foreground Desk, but removing drag handling from the shared Composer also removes file drops from Catch Up replies without providing a replacement owner.
_1 inline comment below._
<sub>Reviewed `af5a55b` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-122-review)</sub>

tella-butler:
Verified the single app-wide drop state at both supported widths.

**Desktop**
Desktop app-wide file drop overlay

**Phone**
Phone app-wide file drop overlay

packages/core/opensession-server/src/frontend/components/Composer.tsx
```diff
@@ -1540,38 +1477,7 @@ export function Composer({
           disabled && "opacity-60",
         )}
         style={surfaceStyle}
-        onDragEnter={handleDragEnter}
-        onDragLeave={handleDragLeave}
-        onDragOver={handleDragOver}
-        onDrop={handleDrop}
```
tella-butler, line 1546: 🟠 **P2** — Preserve file drops in Catch Up replies

`CatchUpDeck` still renders an attachment-capable `Composer` with `images`, `files`, and their change handlers, but it has no replacement drag listener after these shared handlers are removed. On `/catchup`, dropping `report.pdf` onto the Reply composer therefore never reaches `splitAttachments` and the browser default is not prevented. Keep drag ownership out of `Composer`, but add a Catch Up scoped owner that routes the drop to the active reply composer's attachment path.

tella-butler (commented):
OS review · `af5a55b`

PR #125, opened 2026-08-21
Fix compiled binary boot: load the session-kernel worker from an on-disk sidecar

soutar:
## Problem

Every compiled single-executable release is dead on arrival. The server fails to boot with:

    Session kernel actor failed: BuildMessage: ModuleNotFound resolving "/session-kernel-worker.ts" (entry point)

The session kernel is the authoritative owner of session state and starts before the gateway hydrates, so when it can't load the whole server refuses to boot. This affects the v0.4.3/v0.4.4/v0.4.5 artefacts.

## Root cause

The kernel runs in a Worker thread, and the gateway reads its store synchronously by blocking on `Atomics.wait` against a `SharedArrayBuffer` the worker fills. That transport requires a real second thread: it cannot run in-process (Atomics.wait deadlocks a single thread) or as a subprocess (SharedArrayBuffer is not shared across processes).

`bun build --compile` (Bun 1.3.13) does not embed Worker entry points, so `new Worker(new URL("session-kernel-worker.ts", import.meta.url))` resolved at runtime to a bunfs path that was never bundled. Confirmed in isolation: the canonical `new Worker(new URL(...))` pattern works under `bun run` but fails identically under `--compile`, and `new Worker(process.execPath)` tries to parse the binary as source.

## Fix

Ship the worker as a `session-kernel-worker.js` sidecar beside the binary (the same shape as the sharp sidecar) and load it from `process.execPath`'s directory in compiled mode. A source checkout still runs the TypeScript entry, so the Worker stays a real thread and the SharedArrayBuffer transport is unchanged.

- `build-compile.ts` bundles `session-kernel-worker.ts` (`bun build --target=bun`, 72 KB) into the release next to the sharp sidecar, and beside the outfile in `--outfile` mode.
- `actor-runtime.ts` resolves the worker URL: `join(dirname(process.execPath), "session-kernel-worker.js")` when compiled, the `.ts` URL from source.

## Verification

Built the full darwin-arm64 artefact from this branch and installed it through the health-gated update: boots healthy, no rollback. Booting past the health gate means the kernel `hello()` handshake over the Worker + SharedArrayBuffer succeeded, so the actor is genuinely alive. Typecheck clean.

## Known follow-up (not in this PR)

`workflow-worker.ts` and `code-flow-worker.ts` use the same non-embeddable `new Worker(new URL(...))` pattern, so they will fail the same way in a compiled binary when their features run. They are lazy rather than boot-critical, so they do not block the release, but they need the same sidecar treatment.

### DIFF
--- a/docs/compiled-binary.md
+++ b/docs/compiled-binary.md
@@ -12,7 +12,8 @@ compiled in and run in-process. The `--source` install (a git checkout +
 
 ```bash
 # Release artefact (the default install downloads this): the target binary +
-# a sharp sidecar + the service templates + release.json, tarred.
+# a sharp sidecar + the session-kernel worker sidecar + the service templates
+# + release.json, tarred.
 bun scripts/build-compile.ts --os linux --arch arm64 --out ~/.cache/opensession-release
 
 # Just the host binary, for local testing (no sidecar).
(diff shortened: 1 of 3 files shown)


tella-butler:
<details>

Safe to merge. The compiled release now bundles the session-kernel worker as a sidecar, resolves it beside the real executable, and preserves the existing TypeScript worker path for source runs. The release staging and installation layout keep the worker beside the binary as required.
<sub>Reviewed `2d85147` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-125-review)</sub>

</details>

tella-butler:
Safe to merge. The release build stages the bundled session-kernel worker beside the executable, and runtime resolution preserves the source checkout path while loading the sidecar in compiled builds. The latest change consistently passes a string URL to Bun's Worker constructor, and all CI checks pass.
<sub>Reviewed `a75364d` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-125-review)</sub>

PR #126, opened 2026-08-21
Ship every Worker as an on-disk sidecar for compiled builds

soutar:
Follow-up to #125 (session-kernel sidecar). That fix covered one of three Workers; this covers the other two.

## Problem

`workflow-worker.ts` and `code-flow-worker.ts` use the same `new Worker(new URL("./x.ts", import.meta.url))` pattern that `bun build --compile` cannot embed. In a compiled binary they resolve to an unbundled bunfs path and throw `ModuleNotFound` the moment their feature runs — a Workflow tool call, or a code-flow diff analysis. They are lazy (not boot-critical like the kernel), so a compiled release boots fine and then dies on first use of either feature.

Verified empirically by spawning each worker's exact expression in a compiled binary:

```
workflow-worker:  FAIL -> ModuleNotFound resolving "/$bunfs/root/workflow-worker.ts"
code-flow-worker: FAIL -> ModuleNotFound resolving "/$bunfs/root/code-flow-worker.ts"
```

`code-flow.ts`'s `analysisWorkerUrl()` already *tried* a `.js` fallback, but checked `new URL("./code-flow-worker.js", import.meta.url)` — a bunfs path that is always empty — so it fell through to the `.ts` and failed anyway.

## Fix

Factor the compiled/source resolution into a shared `workerEntry(sidecarName, sourceUrl)` helper in `runner-host/exe.ts` and route all three Worker spawns through it (session-kernel included, replacing its inline version from #125). `build-compile.ts` stages all three `.js` sidecars beside the binary from one `WORKER_SIDECARS` list. Compiled mode resolves each from `process.execPath`'s dir; a source checkout runs the TypeScript entry.

Neither lazy worker uses `SharedArrayBuffer`/`Atomics` (unlike the kernel), so a separate thread is not strictly required, but the sidecar keeps all three consistent.

## Verification

- Built with all three sidecars staged; a probe spawning `workflow-worker.js` and `code-flow-worker.js` from beside a compiled binary now loads both with no ModuleNotFound (was the failure above).
- Full darwin-arm64 artefact built and installed through the health-gated update: boots healthy, no rollback (confirms routing session-kernel through the shared helper did not regress boot).
- Typecheck clean; `check-module-side-effects` passes (390 modules, no import-time effects from the new imports).

### DIFF
--- a/packages/core/opensession-server/src/runner-host/exe.ts
+++ b/packages/core/opensession-server/src/runner-host/exe.ts
@@ -60,3 +60,16 @@ export function transcriptSearchWorkerArgv(bun: string, entry: string): string[]
 		? [process.execPath, "transcript-search-worker"]
 		: [bun, entry];
 }
+
+/**
+ * Resolve a Web Worker entry that survives `bun build --compile`, which does not
+ * embed Worker entry points: at runtime `new Worker(new URL("./w.ts",
+ * import.meta.url))` resolves to a bunfs path that was never bundled and fails
+ * with ModuleNotFound. A compiled binary loads the worker from a sibling
+ * `<name>.js` staged beside the executable (scripts/build-compile.ts, next to
+ * the sharp sidecar); a source checkout runs the TypeScript entry at `sourceUrl`.
+ * Keep the sidecar names in sync with build-compile's WORKER_SIDECARS list.
+ */
+export function workerEntry(sidecarName: string, sourceUrl: string | URL): string | URL {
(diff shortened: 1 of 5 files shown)


tella-butler:
Safe to merge. The shared worker resolver preserves source-mode behavior while directing compiled builds to the three staged sidecars, and the build artifact includes each required worker. No correctness, safety, or contract regressions found.
<sub>Reviewed `229d9cc` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-126-review)</sub>

PR #136, opened 2026-08-22
Keep Slack posts attributed to their requester

tella-butler:
## Summary
- add an attributed `slack_post_files` MCP tool so agents no longer need to bypass the per-user Slack connection for media uploads
- run Slack text, replies, and file comments through the session owner's personal token when connected
- identify the requester in bot fallback posts
- instruct sessions never to use shell imports, curl, or host bot tokens for Slack delivery

## Root cause
The session needed to upload a video, but the Slack MCP only supported text. After the editable composer timed out, the agent imported the internal Slack helper from a shell command. That helper used the service bot token, so Slack showed Michael instead of Michiel.

## Validation
- `bun test scripts/mcp-slack.test.ts packages/core/opensession-server/src/server/run-instructions-slack.test.ts`
- `bun run typecheck`
- full suite: 3,745 passed; unrelated environment-dependent failures remain in the existing broad suite

Started by Kent de Bruin in [this OS session](https://os.tella.dev/session/slack-C0BE8KVFBEX-1787392083.169729)

### DIFF
--- a/packages/core/opensession-server/src/server/connections.ts
+++ b/packages/core/opensession-server/src/server/connections.ts
@@ -8,6 +8,7 @@ import { existsSync, readFileSync, copyFileSync, watchFile } from "fs";
 import { writeFileAtomic } from "./shared/atomic-write";
 import { configuredPaths } from "./config";
 import { mcpOauthStatus, mcpSharedGrantHeader, mcpUserGrantHeader, mcpUserGrantToken, oauthPresetFor } from "./mcp-oauth";
+import { resolveTeammate } from "./shared/user-mappings";
 
 const HOME = homeDir();
 // mcp-config.json location. OPENSESSION_MCP_CONFIG env → config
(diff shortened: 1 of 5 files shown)


tella-butler:
Safe once the two P1 findings below are fixed. The PR adds personal-token attribution and Slack file uploads, but the upload tool exposes arbitrary host files and bot fallback attribution can still name the session creator instead of the person requesting the post.
_2 inline comments below._
<sub>Reviewed `241e3c5` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-136-review)</sub>

scripts/mcp-slack.ts
```diff
@@ -268,6 +327,25 @@ async function main(): Promise<void> {
         case "slack_post_message":
           result = await client.postMessage(requiredString(args, "channel_id"), requiredString(args, "text"), options);
           break;
+        case "slack_post_files": {
+          const channel = requiredString(args, "channel_id");
+          const completed = await postSlackFiles(
+            channel,
+            requiredStringArray(args, "file_paths"),
```
tella-butler, line 334: 🔴 **P1** — Restrict uploads to files owned by the current run

`file_paths` is model-controlled and passed directly to `postSlackFiles`, which calls `statSync(path)` and `readFileSync(path)` without containment checks. The stdio MCP process runs as the Open Session service user, so a request such as `/home/ubuntu/.opensession-mcp-oauth.json` can upload the OAuth store, including other users' access and refresh tokens, to Slack. Inject the current run's worktree and scratch roots into this MCP process, resolve each path with `realpath`, and reject anything outside those roots before reading or uploading any file.

packages/core/opensession-server/src/server/connections.ts
```diff
@@ -95,15 +96,37 @@ export function withDynamicCredentials(
...
+            }))
+            .find(({ token }) => !!token);
           const token =
-            candidates
-              .map((u) => mcpUserGrantToken(name, u))
-              .find((t) => !!t) ??
+            personal?.token ??
             mcpSharedGrantHeader(name)?.replace(/^Bearer\s+/i, "");
-          if (token)
+          const actorIdentity = personal?.identity ?? candidates[0];
```
tella-butler, line 108: 🔴 **P1** — Bot fallback attributes the session creator instead of the requester

The candidate order is `[session.startedBy, current prompter]`, as established by `mcpGrantUser: session.startedBy` and `grantUsers = [opts.mcpGrantUser, opts.user]`. When no personal grant exists, `candidates[0]` therefore names the creator. For example, if Kent starts a shared session and Michiel later requests a Slack post while neither has connected Slack, the bot footer says `Sent by Kent` even though Michiel requested it. Keep creator-first token selection, but derive the bot-fallback actor from the current prompter, which is the final candidate in this call path.

tella-butler (commented):
OS review · `241e3c5`

PR #138, opened 2026-08-24
Fix install PATH hint order and honor appOrg for App tokens

soutar:
Two independent first-run fixes, from reviewing a fresh simple-mode install.

## install.sh — PATH hint before the command list

On a fresh install the shell that launched the installer has not picked up the new PATH entry, so `opensession` is not yet resolvable there. The summary printed the `opensession status/doctor/--help` list *above* the `source ~/.zshrc` hint, so a reader who runs the first command hits "command not found" and only then sees the fix. The hint now precedes the command list.

## github-app.ts — honor `appOrg` in installation-owner selection

Installation-token owner selection read only `installationOwner`, which nothing in the setup flow writes. An org App installed on more than one account (e.g. the org plus the owner's personal account) then has no way to disambiguate and throws, dropping to the bot PAT — even though the setup-captured `appOrg` already names the intended owner. Owner selection now falls back `installationOwner → appOrg`, the same precedence `setup-team.ts` already uses.

## Verification
- `bun test` github-auth + runner-bootstrap suites: 33 pass, 0 fail
- `tsc --noEmit`: no new errors (one pre-existing unrelated `pi-ai/bun-oauth` module-resolution error remains)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -899,10 +899,14 @@ if [ "$ADVANCED" != "1" ] && [ "$server_ready" != "1" ]; then
 else
   step "Done"
 fi
+# The PATH hint has to precede the command list: on a fresh install the shell
+# that launched the installer has not picked up the new PATH entry, so a reader
+# who runs `opensession status` from the top of this list before sourcing their
+# profile hits "command not found".
+show_path_refresh_hint
 info "opensession status    ${D}is the server up?${N}"
 info "opensession doctor    ${D}check the install${N}"
 info "opensession --help    ${D}everything else${N}"
-show_path_refresh_hint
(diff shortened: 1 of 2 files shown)


tella-butler:
Safe to merge. The PATH refresh guidance now precedes commands that require it, and GitHub App installation selection correctly falls back from `installationOwner` to the normalized, setup-captured `appOrg`. No correctness, safety, or compatibility issues found.
<sub>Reviewed `1df0d22` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-138-review)</sub>

PR #139, opened 2026-08-24
Reconcile pre-execution creation dead letters

jfrolich:
## Summary
- reconcile only creation branch dead letters produced before physical execution by the retired shared-checkout classifier or empty-base decoder
- run reconciliation through the actor-owned store behind the boot recovery gate
- audit every re-admitted effect and leave unrelated or malformed dead letters untouched
- fix the minimal Codex login environment typing regression exposed by current `main`

## Verification
- `bun run typecheck`
- 77 focused SessionKernel, actor-boundary, and creation-executor tests
- `bun scripts/check-module-side-effects.ts` (409 modules)
- `git diff --check origin/main...HEAD`

### DIFF
--- a/packages/core/opensession-server/src/server/session-kernel/store.ts
+++ b/packages/core/opensession-server/src/server/session-kernel/store.ts
@@ -2376,6 +2376,104 @@ export class SessionKernelStore {
 		return result.changes > 0;
 	}
 
+	/**
+	 * Re-admit branch effects rejected before physical work by compatibility bugs:
+	 * the former shared-checkout classifier and the old empty-base decoder. The
+	 * caller supplies trusted shared destinations; every other failure stays dead.
+	 */
+	retryCompatibleCreationBranchDeadLetters(
+		destinations: ReadonlyArray<{ project: string; worktreePath: string }>,
+		now = Date.now(),
+	): Array<{
+		id: number;
+		sessionId: string;
+		reason: "shared_checkout_destination_adoptable" | "legacy_empty_base_branch";
+	}> {
+		const allowed = new Set(
+			destinations.map(({ project, worktreePath }) =>
+				JSON.stringify([project, worktreePath]),
+			),
+		);
+		const rows = this.db
+			.query(
+				`SELECT id, session_id, payload, last_error
+				 FROM session_kernel_outbox
+				 WHERE kind = 'creation_branch_prepare'
+				   AND dead_lettered_at IS NOT NULL
+				 ORDER BY id
+				 LIMIT 1000`,
+			)
+			.all() as Array<{
+				id: number;
+				session_id: string;
+				payload: string;
+				last_error: string | null;
+			}>;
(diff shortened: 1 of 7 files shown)


tella-butler:
Safe once the two P2 reconciliation gaps below are fixed. The recovery remains actor-owned and boot-gated, but eligible dead letters can be permanently skipped by the bounded scan or by inconsistent path normalization.
_2 inline comments below._
<sub>Reviewed `3e3bb7d` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-139-review)</sub>

tella-butler (commented):
OS review · `3e3bb7d`

packages/core/opensession-server/src/server/session-kernel/store.ts
```diff
@@ -2376,6 +2376,104 @@ export class SessionKernelStore {
...
+			),
+		);
+		const rows = this.db
+			.query(
+				`SELECT id, session_id, payload, last_error
+				 FROM session_kernel_outbox
+				 WHERE kind = 'creation_branch_prepare'
+				   AND dead_lettered_at IS NOT NULL
+				 ORDER BY id
+				 LIMIT 1000`,
```
tella-butler, line 2404: 🟠 **P2** — Unrelated dead letters can permanently starve reconciliation

The query examines only the oldest 1,000 branch dead letters before filtering them in TypeScript. If those rows are unrelated failures, such as 1,000 older `credential unavailable` dead letters, an eligible compatibility false positive with id 1001 is never examined. Every later boot selects the same untouched 1,000 rows, leaving that creation permanently wedged. Keep the bounded batches, but paginate by `id` until all branch dead letters have been inspected.

packages/core/opensession-server/src/server/session-kernel/store.ts
```diff
@@ -2376,6 +2376,104 @@ export class SessionKernelStore {
...
+					payload.branch.length === 0 ||
+					payload.mode !== "adopt_or_create" ||
+					typeof payload.isolated !== "boolean"
+				)
+					continue;
+				const sharedCheckoutFalsePositive =
+					payload.isolated === false &&
+					allowed.has(
+						JSON.stringify([payload.project, payload.worktreePath]),
+					) &&
```
tella-butler, line 2445: 🟠 **P2** — Relative shared-checkout paths never match the normalized allowlist

The caller normalizes configured destinations with `resolve(repo.repo)`, while this comparison uses the historical payload path verbatim. `RepoSection.repo` accepts an arbitrary string and `configuredRepos()` preserves it, so a supported configuration such as `repo: "../opensession"` produced a payload with `worktreePath: "../opensession"`; the allowlist now contains its absolute path and this check rejects the otherwise exact historical error. Normalize `payload.worktreePath` with `resolve` when building the lookup key, matching the current executor's shared-destination comparison.

PR #140, opened 2026-08-24
Prepare creation sandboxes through the actor

jfrolich:
## Summary
- route initial sandbox preparation through the durable creation actor before opening-turn launch
- carry the stable creation identity through web, MCP, and restart recovery paths
- keep the existing run-time ensure only as an idempotent adoption bridge
- fail the actor-owned creation lifecycle when a typed creation effect reaches a permanent dead letter
- fence concurrent opening-create coalescing by creation identity

## Verification
- `bun run typecheck`
- 109 focused SessionKernel, creation-intent, ownership, and effect-executor tests
- `bun scripts/check-module-side-effects.ts` (409 modules)
- `git diff --check`


tella-butler:
### 🤖 OS review

PR #142, opened 2026-08-24
Route a refused WebSocket upgrade to the sign-in card

soutar:
## The lockout

After enabling org GitHub sign-in on a fresh local install, the browser got stuck on **"Connection lost. retrying"** with no way to sign in.

Root cause, traced end to end (server proven by direct probes, client by code):

- When `integrations.github.userPrAuth` is on and the browser holds no accepted session, the server returns **401 on every authenticated route and on the UI WebSocket upgrade** (verified: gate on + no cookie → `/api/auth/status` returns `{required:true, authenticated:false}`, `/ws` → 401, all `/api/*` → 401; `GET /` still 200).
- `useWebSocket` mounts **above** `UserGate`, so the socket 401s and `RestartOverlay` renders "Connection lost. retrying" — a *network* message for what is really an *auth* state.
- The existing recovery (reload into the sign-in gate) fired only when the socket had **previously opened** (`everOpenRef`). A fresh load into a gated instance — the upgrade 401s immediately, never opening — fell through to retry forever, and `UserGate`'s optimistic paint (for a returning browser whose `localStorage` identity survived the reinstall) left the connection overlay standing in for the sign-in card.

This was **not** a server bug: the gate flip and admin roster are correct and atomic, the session cookie is non-`Secure`/`HttpOnly`/`SameSite=Lax` (stores fine on localhost). The defect is that the client surfaces an auth-required 401 as a connection failure.

## The fix

One shared predicate, `authGatesOut(status)` (in `lib/auth-ready.ts`), used by both deciders so they never disagree about what a 401 means:

- **`useWebSocket`**: on a socket that closes without ever opening, publish the gated status instead of reloading. A reload would loop on the sign-in card, whose own socket also 401s. Falls through to keep retrying, so a completed sign-in reconnects without a manual refresh. The previously-open path keeps its reload.
- **`UserGate`**: render the GitHub sign-in card the moment a gated status is published, overriding the optimistic paint.

Net: a gated browser with no session always lands on the sign-in card, never on an infinite reconnect overlay.

## Why this is the fix, not "don't enable sign-in for localhost"

Per-user sign-in is a legitimate mode for a single-machine install: run it on a VPS, `opensession bind` it onto a tailnet, and share the URL with colleagues, and the gate is exactly what protects it. So the right behavior is a gate that recovers cleanly — the admin who enables it and every colleague who opens the URL land on the sign-in card — which is what this PR does.

## Verification
- Reproduced the server side deterministically on a real install (flipped `userPrAuth` on, probed un-cookied endpoints, restored) — the 401s above are measured, not assumed.
- `bun test lib/auth-ready.test.ts` — 4 pass (the shared predicate's contract).
- `tsc --noEmit` — no new errors.
- **No screenshot**: the change swaps a connection overlay for the existing sign-in card; exercising it live needs a real GitHub device flow. I can build a patched binary and walk it on the affected Mac if you want visual proof before merge.

### DIFF
--- a/packages/core/opensession-server/src/frontend/components/UserPicker.tsx
+++ b/packages/core/opensession-server/src/frontend/components/UserPicker.tsx
@@ -11,7 +11,7 @@ import { cn } from "../ui/cn";
 import { DeviceCode } from "../ui/device-code";
 import { InlineAlert } from "../ui/state";
 import { PulseDot } from "../ui/status";
-import { AUTH_STATUS_EVENT } from "../lib/auth-ready";
+import { AUTH_STATUS_EVENT, authGatesOut } from "../lib/auth-ready";
 
 /**
  * Mutable compatibility view for older consumers. `usePeople()` owns the
@@ -85,6 +85,14 @@ let authStatusCache: AuthStatus | null = null;
 function setAuthStatusCache(status: AuthStatus) {
   authStatusCache = status;
   window.dispatchEvent(new Event(AUTH_STATUS_EVENT));
+}
+
+/** Publish an auth status discovered outside UserGate — the WebSocket layer
+ *  learns the gate turned on (or that a fresh load landed on a gated instance)
+ *  from a refused upgrade, and drives UserGate to the sign-in card through
+ *  this, without a reload that would loop on the card's own refused socket. */
+export function publishAuthStatus(status: AuthStatus): void {
+  setAuthStatusCache(status);
 }
 
 /** Reactive sign-in state; null until /api/auth/status answers (or when the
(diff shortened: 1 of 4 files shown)


tella-butler:
Safe to merge. The shared predicate correctly identifies an unauthenticated gated browser, the refused-upgrade path publishes that state without creating a reload loop, and UserGate replaces the optimistic app paint and reconnect UI with the existing sign-in card. The server payload, mount order, retry behavior, and post-sign-in cache update remain consistent.
<sub>Reviewed `439cd51` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-142-review)</sub>

PR #144, opened 2026-08-24
Settle durable creation opening lifecycle

jfrolich:
## Summary
- cross the actor-owned creation lifecycle into `opening_dispatched` immediately before each host, sandbox, or Runner opening launch
- settle successful openings as `ready` and setup failures as `failed`
- fail closed if an opening is attempted while a durable preparation effect is still pending
- make terminal lifecycle settlement idempotent for replay and restart paths

This is a bounded creation-cutover slice. Opening launch itself still uses the existing adapters and will move behind the durable `creation_opening_turn` effect in the next dependency-ordered slice.

## Verification
- `bun test packages/core/opensession-server/src/server/session-kernel/creation-intents.test.ts packages/core/opensession-server/src/server/session-kernel/kernel.test.ts packages/core/opensession-server/src/server/session-kernel/ownership.test.ts packages/core/opensession-server/src/server/session-create-plan.test.ts` (84 tests, 868 assertions)
- `bun run typecheck`
- `bun scripts/check-module-side-effects.ts` (407 modules)
- `git diff --check`

### DIFF
--- a/packages/core/opensession-server/src/server/session-create.ts
+++ b/packages/core/opensession-server/src/server/session-create.ts
@@ -81,10 +81,13 @@ import { resolveWorkspaceModelPreset } from "./workspace-model-presets";
 import { type Workspace, getWorkspace, updateWorkspace, } from "./workspaces";
 import {
 	ensureCreationPlanned,
+	markCreationOpeningDispatched,
 	requestCreationBranch,
 	requestCreationCredential,
 	requestCreationSandbox,
 	requestCreationWorkspace,
+	settleCreationFailed,
+	settleCreationSucceeded,
 } from "./session-kernel";
 import { AUTO_REPO, ensureAskCheckout, ensureScratchDir, getRepo, isRegisteredWorktree, listWorktrees, NO_REPO, repoForPath, repoForPathOrNull, resolveUniqueBranch, sharedCheckoutForNewSessions, worktreeHeadBranch, worktreePathFor, } from "./worktree";
 import { type WSClientData, broadcastToSession, preparingWorkspaces, } from "./ws-hub";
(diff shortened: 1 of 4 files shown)


tella-butler:
Safe once the two P1 lifecycle gaps below are fixed. The PR records opening dispatch and terminal outcomes on the normal path, but successful openings can be misclassified by follow-up work, and journal-based restart recovery never settles the durable creation lifecycle.
_2 inline comments below._
<sub>Reviewed `76893d7` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-144-review)</sub>

packages/core/opensession-server/src/server/session-create.ts
```diff
@@ -775,6 +783,7 @@ export async function openCreatedSession(
 						?.map((repo) => repo.dir)
 						.filter(Boolean),
 				}, { timeoutMs: 15 * 60_000 });
+				markCreationOpeningDispatched(bksId, creationIdentity);
```
tella-butler, line 786: 🔴 **P1** — Recovered openings never settle the creation lifecycle

After this transition, the sandbox launcher durably journals the run and the global journal listener acknowledges the create prompt dispatch. If the server restarts while that opening is running, boot recovery resumes the journal-owned run, but `resumePlannedCreate` is skipped because the create dispatch is gone. No run-journal recovery path calls either new settlement helper, so even when the recovered opening completes successfully, this session remains `opening_dispatched` permanently. The same gap applies to failures recovered through the run journal. Have recovered opening runs settle the associated creation state from their terminal outcome, or retain another creation-specific recovery record until terminal settlement.

packages/core/opensession-server/src/server/session-create.ts
```diff
@@ -1103,6 +1115,7 @@ export async function openCreatedSession(
 			else
 				onHumanAsksSessionIdle(bksId);
 		}
+		settleCreationSucceeded(bksId, creationIdentity);
```
tella-butler, line 1118: 🔴 **P1** — Queued follow-up failures mark a successful creation as failed

This success transition runs only after the web create path drains every prompt queued during the opening. For example, if the opening completes successfully, then a queued follow-up cannot start because its Runner disconnected, `drainQueue` rethrows the launch error and the outer catch records the creation as `failed`. That failure belongs to a later turn, not session setup, and future creation intents will now reject the otherwise valid session as already failed. Settle the creation as `ready` immediately after the opening turn's outcome is persisted, before handing off to queue draining, and keep later-turn errors outside the creation-failure catch.

tella-butler (commented):
OS review · `76893d7`

PR #147, opened 2026-08-24
Filter model choices by configured providers

jfrolich:
## Summary
- hide the Dial until both OpenAI and Anthropic accounts are configured
- hide workspace and global presets when any required provider is unavailable
- keep current models such as Fable and Sol first class while reserving Legacy for retired direct-SDK entries

## Verification
- `bun test packages/core/opensession-server/src/server/model-catalog.test.ts packages/core/opensession-server/src/frontend/components/ModelEffortSelect.models.test.ts packages/core/opensession-server/src/server/models.test.ts packages/core/opensession-server/src/frontend/lib/model-engine.test.ts`
- `bun run typecheck`
- desktop and phone visual captures against an isolated OpenAI-only dev instance

The complete repository test command was also run. It remains red in unrelated existing suites, including shared DOM setup, sandbox prewarm timeouts, and the workspace runtime v7/v8 expectation.

Started by Jaap Frolich in [this Assistant session](http://100.81.254.102:3850/session/os-01a03329-7683-78c8-95f3-bda503724283)

### DIFF
--- a/packages/core/opensession-server/src/server/routes/models.ts
+++ b/packages/core/opensession-server/src/server/routes/models.ts
@@ -8,7 +8,10 @@
 
 import { requestUser, type RouteContext } from "./context";
 import { DIAL_PRESETS, KNOWN_MODELS, ORCHESTRATOR_PRESETS, accountProviderForModel, getDefaultModel, getModelFallbackAuto, interactiveDefaultModel, modelEfforts, piModelLabel, refreshPickerModels, setDefaultModel, setInteractiveDefaultModel, setModelFallbackAuto, toPiModel } from "../models";
-import { orchestratorEnabled } from "../model-providers";
+import { modelProviders, orchestratorEnabled } from "../model-providers";
+import { modelUpstreamProvider, presetFitsConfiguredProviders } from "../model-catalog";
+import { listAccountsPublic } from "../claude-accounts";
+import { listCodexAccountsPublic } from "../codex-accounts";
 import { type Sandbox } from "../sandbox";
 import { suggestBranchName } from "../suggest-branch";
 import { suggestRepos } from "../suggest-repos";
@@ -36,6 +39,15 @@ export async function handleModelsRoutes(
 			? getWorkspace(url.searchParams.get("workspace")!)
 			: null;
 		const settings = workspaceModelSettings(workspace);
+		// Temporary exhaustion does not change the catalog. An account existing is
+		// enough; only adding or removing provider capacity changes what is offered.
+		const configuredProviders = new Set<string>([
+			...(listAccountsPublic().length ? ["anthropic"] : []),
+			...(listCodexAccountsPublic().length ? ["openai"] : []),
+			...Object.entries(modelProviders())
+				.filter(([, provider]) => !!provider.apiKey)
+				.map(([provider]) => provider),
+		]);
 		// One engine-agnostic list: every entry (models and presets alike) runs
 		// on any configured engine — the composer's Engine choice routes it by
 		// prefix at dispatch (`engines` below is the only engine signal). Native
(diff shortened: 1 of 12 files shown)


tella-butler:
Safe once the unavailable-default handling is fixed. The provider filtering generally matches the stated goal, but a persisted default can now disappear from the catalog while remaining the model used for untouched new sessions. Provider changes can also leave the cached picker stale.
_2 inline comments below._
<sub>Reviewed `c5067d7` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-147-review)</sub>

packages/core/opensession-server/src/server/routes/models.ts
```diff
@@ -46,7 +58,11 @@ export async function handleModelsRoutes(
 		const engineModels = KNOWN_MODELS.filter((m) => m.provider === "pi");
 		const engineConfigured = engineModels.length > 0;
 		const visibleModels = (engineConfigured ? engineModels : KNOWN_MODELS)
-			.filter((model) => model.group !== "dial" && model.group !== "orchestrator");
+			.filter((model) => model.group !== "dial" && model.group !== "orchestrator")
+			.filter((model) => {
+				const provider = modelUpstreamProvider(model.id);
+				return !!provider && configuredProviders.has(provider);
```
tella-butler: 🔴 **P1** — An unavailable default remains active after its picker row is removed

The filtering removes models whose provider is unavailable, but the response still returns `interactiveDefaultModel()` unchanged. For example, set Fable as the default, remove the last Anthropic account, and leave an OpenAI account configured: `/api/models` contains only OpenAI models but still returns Fable as `default`. `preferredNewSessionModel()` then returns `""` for an untouched composer, and `session-create.ts` interprets that as `interactiveDefaultModel()`, creating the session on Fable. `engineStatus()` explicitly treats that state as blocked because no Claude account is available. Ensure the effective default belongs to the filtered catalog and make the no-model session creation path use that same capacity-aware default, rather than merely hiding its row.

packages/core/opensession-server/src/server/routes/models.ts
```diff
@@ -36,6 +39,15 @@ export async function handleModelsRoutes(
 			? getWorkspace(url.searchParams.get("workspace")!)
 			: null;
 		const settings = workspaceModelSettings(workspace);
+		// Temporary exhaustion does not change the catalog. An account existing is
+		// enough; only adding or removing provider capacity changes what is offered.
+		const configuredProviders = new Set<string>([
+			...(listAccountsPublic().length ? ["anthropic"] : []),
```
tella-butler: 🟠 **P2** — Account changes do not invalidate the provider-dependent model catalog

`fetchModels()` caches each catalog for 60 seconds and returns the cached value even when starting its background refresh. The account add/remove handlers do not call `invalidateModelsCache()`. Consequently, if `/new` cached a catalog while both providers existed and the last Anthropic account is then removed in Settings, navigating back can still show and select the Dial or Anthropic models. Even after the TTL, the first fetch renders the stale value and does not update the mounted caller when the background request finishes. Invalidate all model catalogs after successful account/provider capacity mutations, or stop caching this provider-dependent response.

tella-butler (commented):
OS review · `c5067d7`

packages/core/opensession-server/src/server/routes/models.ts
jfrolich: Fixed in 9ee1228fb: the catalog and both untouched session-creation paths now share a capacity-aware default resolver, preserving preset ids and falling back to the first configured model when the persisted default is unavailable.

packages/core/opensession-server/src/server/routes/models.ts
jfrolich: Fixed in 9ee1228fb: settled model catalogs are no longer cached client-side. Sequential picker loads now fetch fresh provider-dependent catalogs, while the existing request helper still deduplicates only concurrent GETs.

PR #148, opened 2026-08-24
Keep remote sandbox path independent of release checkout

jfrolich:
## Summary
- keep the checkout path inside remote sandboxes fixed at `/home/ubuntu/projects/opensession`
- prevent the host service's dedicated release-checkout path from leaking into remote bootstrap commands
- add a regression assertion for the guest path contract

## Production impact
After moving the service to `/home/ubuntu/releases/opensession`, a durable Daytona preparation effect repeatedly attempted to chmod that host-only path inside the guest. The guest checkout remains at `/home/ubuntu/projects/opensession`.

## Verification
- `bun test packages/core/opensession-server/src/server/sandbox/adapters/bootstrap-install.test.ts` (4 tests)
- `bun run typecheck`
- `bun scripts/check-module-side-effects.ts` (407 modules)
- `git diff --check`

### DIFF
--- a/packages/core/opensession-server/src/server/sandbox/adapters/bootstrap.ts
+++ b/packages/core/opensession-server/src/server/sandbox/adapters/bootstrap.ts
@@ -148,7 +148,9 @@ const REMOTE_NODE_MAJOR = Number(REMOTE_NODE_VERSION.split(".")[0]);
 const REMOTE_JUST_VERSION = "1.43.1";
 const REMOTE_GH_VERSION = "2.83.1";
 const REMOTE_RUNTIME_REVISION = "workspace-runtime-v8";
-const REMOTE_REPO = REPO_ROOT; // /home/ubuntu/projects/opensession
+// This path exists inside every remote sandbox. It must not inherit the host
+// service's checkout path (for example a dedicated production release tree).
+export const REMOTE_REPO = `${REMOTE_HOME}/projects/opensession`;
 export const REMOTE_RUNNER_BINARY = `${REMOTE_HOME}/.local/bin/opensession-runner`;
 const BOOTSTRAP_MARKER = `${REMOTE_HOME}/.bks-bootstrapped`;
 /** Where per-launch openai seed material lands in-sandbox — threaded to the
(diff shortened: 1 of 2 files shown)


tella-butler:
Safe once the P2 below is fixed. The bootstrap checkout now correctly uses the stable guest path, but the explicitly supported source fallback still embeds the host release checkout and can fail to start a run.
_1 inline comment below._
<sub>Reviewed `27641c1` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-148-review)</sub>

packages/core/opensession-server/src/server/sandbox/adapters/bootstrap.ts
```diff
@@ -148,7 +148,9 @@ const REMOTE_NODE_MAJOR = Number(REMOTE_NODE_VERSION.split(".")[0]);
 const REMOTE_JUST_VERSION = "1.43.1";
 const REMOTE_GH_VERSION = "2.83.1";
 const REMOTE_RUNTIME_REVISION = "workspace-runtime-v8";
-const REMOTE_REPO = REPO_ROOT; // /home/ubuntu/projects/opensession
+// This path exists inside every remote sandbox. It must not inherit the host
+// service's checkout path (for example a dedicated production release tree).
+export const REMOTE_REPO = `${REMOTE_HOME}/projects/opensession`;
```
tella-butler, line 153: 🟠 **P2** — Use the guest checkout for the runner source fallback too

`remoteRunnerHostCommand()` still constructs its fallback from `HOST_ENTRY`, which is derived from the host-side `REPO_ROOT`. With the service running from `/home/ubuntu/releases/opensession`, a missing or non-executable compiled runner makes the guest execute `/home/ubuntu/releases/opensession/packages/core/opensession-server/src/runner-host/host.ts`, even though its checkout is now explicitly `/home/ubuntu/projects/opensession`. That file does not exist in the guest, so the documented and tested source fallback cannot start the run. Build the fallback entry path from `REMOTE_REPO` instead, and tighten the existing fallback test to assert the complete guest path.

tella-butler (commented):
OS review · `27641c1`

PR #153, opened 2026-08-24
Run the GitHub PR agent on the App installation token

soutar:
First stage of making the **GitHub App the single credential** (retiring the bot PAT). It moves the PR agent — the primary consumer — onto the App installation token.

## Why
The PR agent authenticated only with `GITHUB_API_TOKEN`. A configured GitHub App is **strictly more capable**: installation-scoped (out-of-org writes still fail at GitHub's side, per security-model.md §GitHub credential scoping), its own 5000/hr bucket, and `checks:read` the PAT can't have. But nothing routed the agent's writes through it, so an install with an App and no PAT left the agent idle.

## What
- **Scoped mint** (`github-app.ts`): `githubAppInstallationToken({ write })` mints a write token (`pull_requests`/`issues`/`contents: write`, still installation-scoped) beside the existing read-only one, cached per scope.
- **One resolver**: `githubToken({ write })` → App token when configured, else `GITHUB_API_TOKEN`. `githubAppConfigured()` is the sync precondition.
- **Agent routed** (`github-rest.ts`): REST + GraphQL resolve the credential per call (installation tokens are short-lived, unlike a static PAT). `githubConfigured()` is App-aware.
- **Identity** (`config.ts`): the App posts as `<slug>[bot]`; `githubBotLogins()` lists it so the agent recognises its own App-posted comments and never answers itself.

## Verification
- `bun test` github/config suites: pass (2 pre-existing `shipped-change-notify` screenshot failures are unrelated and present on `main`).
- New test pins the `<slug>[bot]` identity recognition.
- `tsc --noEmit`: no new errors.
- **Not live-validated**: real App writes need an installed App posting to real PRs — that needs the operator's environment, so this is logic-tested only. I can walk a live check on the affected Mac.

## Remaining stages (this migration)
1. **Peripheral PAT readers** → App: rate-limit probe (`github-limit.ts`), clones (`sandbox/bootstrap.ts`), `pr-images`, `account-health`, Slack `github-reviews`.
2. **Webhooks on the App**: retire `GITHUB_WEBHOOK_SECRET` + `gh webhook forward` for the App's native webhook config (fits the VPS-on-a-tailnet topology — App delivers to the bound URL).
3. **Status unification**: "GitHub configured" = App installed; one `githubStatus()` so the setup/Connections surfaces can't disagree (the visible "off vs added"). The single card itself is a UX change (Kent's).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Safe rollout

[...]

PR #157, opened 2026-08-24
Grant the App the right permissions and let it hold a private key

soutar:
Stage (b) of making the GitHub App the single credential. **Stacked on #153** — review/merge that first.

The stage-1 live validation (agent posting on the App token) exposed two concrete gaps in the **existing device-flow App setup**. This keeps that working flow and just completes it — no new create path.

## Gap 1 — permission drift
The two App-create builders under-requested and disagreed: Connections omitted `issues`, **neither requested `checks`**. So `github-app.ts` minted scopes the App was never granted, every installation token 422'd, and it fell back to the bot PAT. The `tellahq` App in testing was granted only `contents:write, pull_requests:write, members:read, metadata:read` — missing exactly `issues` and `checks`.

**Fix:** one canonical definition (`shared/github-app-permissions.ts`) — a grant set the create URL requests, with the read and write mints as strict subsets. A test asserts every mint scope is within the grant (a mint is all-or-nothing, so drift = silent PAT fallback). Both create builders and both mints read it.

## Gap 2 — no private key
Installation tokens sign a JWT with the App's private key, but the device-flow setup captured client id/slug/secret and never the key — so the App could never mint installation tokens.

**Fix:** the key is a one-click **Generate a private key** in the App's settings UI. `POST /api/connections/github/app` now accepts an optional `privateKey` and stores the pasted PEM at `KEY_PATH` via `writeGithubAppKey` (validated as a PEM; refused when a key is env-managed). Nothing else about the flow changes — **device flow stays on for sign-in**, so there is no device-flow tradeoff.

## Scope / ownership
- **Ours (here):** canonical permissions, both builders + mints aligned, the private-key field + storage.
- **Kent (UI):** a "Generate a private key" link + a paste field on the create-app card.

## Verification
- `bun test` permission-invariant + connections + setup-auth suites: pass (setup-auth assertion updated for the now-correct `checks`/`statuses` params).
- `tsc --noEmit` + module side-effects: clean.
- **Live-proven upstream:** with the two missing perms added to the real App and the key in place, the actual agent code minted a write token and posted a comment as `…[bot]` with no PAT (the stage-1 validation).

## Note
An earlier revision of this PR added a GitHub App *manifest* create flow to auto-return the key. That was over-engineered — the existing flow works and the key is a paste away — so it was dropped in favor of the private-key field above.

## Safe rollout

Provisioning the App and private key does not switch live bot traffic. The separate bot-credential setting remains on PAT until an operator deliberately enables the App.

Created by [this OS session](https://os.tella.dev/session/os-01a033fd-020d-76bc-be85-60b19b241fbc)


tella-butler:
<details>

Safe once the P1 findings below are fixed. The permission centralization is sound, but the manifest-created App omits a setting required by the only user authentication flow, and provisioning can overwrite or retain credentials from a different App. Webhook-secret persistence can also fail while the callback reports success.
_4 inline comments below._
<sub>Reviewed `e3ae466` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-157-review)</sub>

</details>

packages/core/opensession-server/src/server/routes/connections.ts
```diff
@@ -743,6 +754,130 @@ export async function handleConnectionsRoutes(
...
+		const manifest = {
+			name: `Open Session (${randomBytes(2).toString("hex")})`,
+			url: base,
+			// Native webhook delivery is a later migration; keep it inactive so the
+			// existing gh-forward path still governs. The conversion still returns a
+			// webhook secret, stored for when native delivery is turned on.
+			hook_attributes: { url: `${base}/github/webhook`, active: false },
+			redirect_url: `${base}/api/connections/github/app/manifest/callback`,
+			public: false,
+			default_permissions: GITHUB_APP_GRANT_PERMISSIONS,
```
tella-butler: 🔴 **P1** — The manifest does not enable the required Device Flow

The new manifest includes permissions and events but omits `device_flow_enabled`. GitHub Apps are created with Device Flow disabled unless it is explicitly enabled; this codebase confirms that it is mandatory (`github-auth.ts` maps GitHub's `device_flow_disabled` response because device flow is the only connection/sign-in mechanism), and both existing create-URL builders explicitly send `device_flow_enabled=true`. A user completing this new create → install flow will therefore reach Connect and be rejected before receiving a device code. Include the corresponding manifest setting if GitHub supports it, or add a mandatory settings step before redirecting to installation and verify it in the browser round-trip.

packages/core/opensession-server/src/server/routes/connections.ts
```diff
@@ -743,6 +754,130 @@ export async function handleConnectionsRoutes(
...
+		} | null;
+		if (!conv.ok || !app?.client_id || !app.pem) {
+			console.warn(`[github-app] manifest conversion failed (${conv.status})`);
+			return new Response(
+				"Could not finish creating the GitHub App. Start again from Settings.",
+				{ status: 502 },
+			);
+		}
+		const { writeGithubAppKey } = await import("../github-app");
+		writeGithubAppKey(app.pem);
```
tella-butler: 🔴 **P1** — Manifest provisioning can corrupt an env-managed App setup

Unlike the sibling POST/DELETE App routes, this flow never rejects `githubAppConfigSource() === "env"`. With `OPENSESSION_GITHUB_CLIENT_ID=Iv1.old`, the callback stores the new App's key and config client ID, but `githubUserAuthSettings()` continues using the old env client ID, so JWTs are signed with a key belonging to another App and every installation-token mint fails. If `OPENSESSION_GITHUB_APP_KEY` is also set, `writeGithubAppKey()` instead throws after the one-time conversion code and CSRF state have been consumed, leaving the newly created App unprovisioned and the request as an unhandled 500. Refuse manifest creation when the App identity or key is env-managed, matching the existing App routes, and recheck that invariant before consuming the callback.

packages/core/opensession-server/src/server/routes/connections.ts
```diff
@@ -743,6 +754,130 @@ export async function handleConnectionsRoutes(
...
+		const { rawConfig, persistRawConfig, withConfigMutationLock } = await import(
+			"../config-mutation"
+		);
+		await withConfigMutationLock(async () => {
+			const config = rawConfig();
+			const github = githubIntegrationSection(config, true)!;
+			github.oauthClientId = app.client_id;
+			if (app.client_secret) github.oauthClientSecret = app.client_secret;
+			if (app.slug) github.appSlug = app.slug;
+			if (appOrg) {
```
tella-butler: 🔴 **P1** — Creating a personal App retains stale organization intent

This branch sets `appOrg` and `authOnConnect` for an organization-owned App but has no personal-owner `else`, unlike the existing App POST route. For example, replacing an unconsumed `acme-inc` org App with a personal App leaves `appOrg: "acme-inc"` and `authOnConnect: true`. `githubAppInstallationToken()` then searches the personal App's installations for `acme-inc` and rejects the valid installation, while the next device connection can incorrectly switch the instance into team sign-in mode. Clear `appOrg`, `authOnConnect`, and the same stale forwarding fields cleared by the sibling POST path when `appOrg` is empty.

packages/core/opensession-server/src/server/routes/connections.ts
```diff
@@ -743,6 +754,130 @@ export async function handleConnectionsRoutes(
...
+			}
+			persistRawConfig(config);
+			return null;
+		});
+		if (app.webhook_secret) {
+			const { applyEnvFileEdits } = await import("../env-file-edit");
+			try {
+				applyEnvFileEdits({ GITHUB_WEBHOOK_SECRET: app.webhook_secret });
+			} catch (e) {
+				console.warn(`[github-app] could not store webhook secret: ${String(e).slice(0, 120)}`);
```
tella-butler: 🟠 **P2** — Webhook-secret write failures are reported as successful provisioning

The conversion returns the webhook secret only once, but an `applyEnvFileEdits` failure is swallowed and the browser is redirected as though setup completed. A supported configuration such as a writable `OPENSESSION_CONFIG` alongside a read-only `OPENSESSION_ENV_FILE=/etc/opensession/env` stores the App key and config, loses the returned secret, and gives the operator no actionable failure. Native webhook delivery later cannot verify this App's signatures without manually regenerating and installing another secret. Treat storing the secret as a required provisioning step and return an actionable failure rather than consuming it silently.

tella-butler (commented):
OS review · `e3ae466`

tella-butler:
<details>

Safe once the three P1 provisioning failures below are fixed. The latest commit correctly clears stale personal-App intent and reports webhook-secret write failures, but organization onboarding still cannot complete, environment overrides remain insufficiently guarded, and partial persistence can destroy a working App credential.
_3 inline comments below._
<sub>Reviewed `daaf402` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-157-review)</sub>

</details>

packages/core/opensession-server/src/server/routes/connections.ts
```diff
@@ -743,6 +754,178 @@ export async function handleConnectionsRoutes(
...
+		const org = (url.searchParams.get("org") || "").trim();
+		const base = configuredServer().publicBaseUrl.replace(/\/$/, "");
+		const state = randomBytes(16).toString("hex");
+		manifestStates().set(state, { org, expiresAt: Date.now() + 15 * 60_000 });
+		// A GitHub App manifest cannot enable Device Flow (there is no manifest
+		// field for it; the create-URL builders pass device_flow_enabled=true as a
+		// form param, which manifests do not accept). This flow provisions the
+		// BOT — the private key for installation tokens — which needs no device
+		// flow. Per-user sign-in DOES rely on device flow, so a manifest App used
+		// for sign-in still needs Device Flow enabled by hand in the App settings.
```
tella-butler: 🔴 **P1** — The manifest flow still cannot complete organization sign-in setup

The new comment confirms the manifest cannot enable Device Flow, but the callback still sets `authOnConnect = true` for an organization-owned App. For an App created for `acme`, the subsequent connection calls `startGithubDeviceFlow()`, which this codebase explicitly handles as `device_flow_disabled`; no account is connected, so `bootstrapUserAuthOnConnect()` never runs and operator-mode setup cannot complete. A future UI warning does not make the current backend flow functional. Either keep manifest provisioning explicitly bot-only and do not set `authOnConnect`, or add an enforced manual Device Flow step before recording that sign-in intent.

packages/core/opensession-server/src/server/routes/connections.ts
```diff
@@ -743,6 +754,178 @@ export async function handleConnectionsRoutes(
...
+		if (webAuthRequired())
+			return Response.json(
+				{ error: "Not available while GitHub sign-in is on" },
+				{ status: 403 },
+			);
+		// Match the sibling App routes: never provision over an env-managed App,
+		// where githubUserAuthSettings() would keep using the old env client id and
+		// sign JWTs against a key that belongs to a different App.
+		const { githubAppConfigSource } = await import("../github-auth");
+		if (githubAppConfigSource() === "env")
```
tella-butler: 🔴 **P1** — The env-managed App guard only checks the client ID

The new guard does not fully protect environment-managed setups because `githubAppConfigSource()` only checks `OPENSESSION_GITHUB_CLIENT_ID`. The codebase independently supports `OPENSESSION_GITHUB_CLIENT_SECRET`, `OPENSESSION_GITHUB_APP_SLUG`, and `OPENSESSION_GITHUB_APP_KEY`, all of which override or constrain the newly persisted values. For example, with a config-managed client ID and an old `OPENSESSION_GITHUB_CLIENT_SECRET`, provisioning writes the new client ID while refreshes continue using the old secret, causing every user token to fail renewal after roughly eight hours. With `OPENSESSION_GITHUB_APP_KEY`, the route creates and converts the GitHub App before `writeGithubAppKey()` throws, leaving an orphaned App. Reject the flow on both GET and callback whenever any App credential/identity environment override is active.

tella-butler (commented):
OS review · `daaf402`

tella-butler:
<details>

Safe once private-key provisioning is connected to the shipped setup flow. Permission grants and mint scopes are now aligned, but the replacement for the removed manifest flow neither collects the key nor safely validates it before overwriting an existing credential.
_2 inline comments below._
<sub>Reviewed `3369b1d` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-157-review)</sub>

</details>

packages/core/opensession-server/src/server/routes/connections.ts
```diff
@@ -772,6 +773,12 @@ export async function handleConnectionsRoutes(
 		// Present ⇒ the App is owned by an org (owner=Organization); empty/absent ⇒
 		// a personal App (single-user).
 		const appOrg = typeof body?.appOrg === "string" ? body.appOrg.trim() : "";
+		// The App's private key (PEM), generated once in the App's settings UI.
+		// Optional here — an App can be configured for per-user sign-in (device
+		// flow, keyless) alone — but it is what lets installation tokens mint, so
+		// without it the bot/agent and checks-read stay on the PAT.
+		const privateKey =
+			typeof body?.privateKey === "string" ? body.privateKey.trim() : "";
```
tella-butler, line 781: 🔴 **P1** — The setup flow never sends the new private-key field

`privateKey` is accepted here, but the only shipped caller in `Connections.tsx:1112-1117` still posts `{ clientId, slug, secret, appOrg }`. The wizard also explicitly tells operators to ignore GitHub's private-key banner. Consequently, completing the normal setup still creates no key file, `githubAppInstallationToken()` returns `null`, and bot/check operations continue falling back to the PAT, so this PR no longer closes the stated private-key gap after removing the manifest flow. Wire a generated-key input through the wizard and include it in this request, or restore another end-to-end provisioning flow that supplies the key.

packages/core/opensession-server/src/server/routes/connections.ts
```diff
@@ -781,6 +788,24 @@ export async function handleConnectionsRoutes(
 				{ error: "clientId, slug and secret are required" },
 				{ status: 400 },
 			);
+		if (privateKey && !/-----BEGIN [A-Z ]*PRIVATE KEY-----/.test(privateKey))
```
tella-butler: 🟠 **P2** — Header-only validation can overwrite a working key with invalid PEM

This regex accepts values such as `-----BEGIN PRIVATE KEY-----\ntruncated`, after which `writeGithubAppKey()` immediately replaces the existing credential and the endpoint reports success. The next JWT signing attempt fails inside `createSign`, silently returning to the PAT, while the prior working key has been lost. Parse the supplied PEM with `createPrivateKey` or perform an equivalent signing validation before writing it; reject invalid or unsupported keys without touching the existing file.

tella-butler (commented):
OS review · `3369b1d`

tella-butler:
<details>

Safe once the remaining P1 is fixed. The permission sets are now centralized and pasted keys are parsed before storage, but the guided setup still never captures or submits the private key, so the stated installation-token path remains unavailable through the product.
_1 inline comment below._
<sub>Reviewed `e01aaee` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-157-review)</sub>

</details>

tella-butler:
<details>

### 🤖 OS review

</details>

packages/core/opensession-server/src/server/routes/connections.ts
soutar: Looks addressed in 138ee68fe and ae0a8d41e: both shipped GitHub App setup surfaces now collect the private key and submit it to their respective setup routes; 2d8c04b11 carries those fixes forward with the explicit credential cutover.

tella-butler:
Safe once the credential-lifecycle issues below are fixed. The permission sets and both setup surfaces are now aligned, and the previous manifest-flow blockers are gone. However, removing or unsuccessfully updating an App can leave the private key and cached credentials in an incorrect state.
_3 inline comments below._
<sub>Reviewed `28c0ea0` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-157-review)</sub>

packages/core/opensession-server/src/server/routes/connections.ts
```diff
@@ -781,6 +788,42 @@ export async function handleConnectionsRoutes(
...
+					{ error: "privateKey must be a valid PEM private key" },
+					{ status: 400 },
+				);
+		}
+		// Store the key first: if it is env-managed writeGithubAppKey refuses, and
+		// we surface that before touching config rather than half-configuring.
+		if (privateKey) {
+			const { writeGithubAppKey } = await import("../github-app");
+			try {
+				writeGithubAppKey(privateKey);
```
tella-butler, line 819: 🔴 **P1** — Removing an App leaves its managed key and cached tokens active

This route now owns writing the App key, but `DELETE /api/connections/github/app` only clears the config fields. For example, after App A has minted a token, removing its configuration leaves both `~/.opensession-github-app.pem` and the global token caches intact; `githubAppInstallationToken()` returns a cached token before checking the now-cleared client ID, so bot operations can continue as App A for up to the token cache lifetime. Creating App B without pasting a key then also pairs B's client ID with A's stale key, while `githubAppConfigured()` incorrectly reports it configured based on file existence. Add a managed-key removal helper that deletes the non-env-managed key and clears both caches, and invoke it from the App DELETE path.

packages/core/opensession-server/src/server/routes/setup.ts
```diff
@@ -379,14 +380,54 @@ export async function handleSetupRoutes(
...
       body.botCredential === undefined
     ) {
       return Response.json({ error: "Nothing to change" }, { status: 400 });
     }
+    // Store the key first: writeGithubAppKey refuses an env-managed key, and we
+    // surface that before writing config rather than half-configuring.
+    if (privateKey) {
+      const { writeGithubAppKey } = await import("../github-app");
+      try {
+        writeGithubAppKey(privateKey);
```
tella-butler, line 423: 🔴 **P1** — A rejected settings update can still replace the working App key

The key is persisted before entering the config mutation that can still reject the request. A concrete path is submitting a new key together with `userPrAuth: true` on an install with no connected account or team member: `writeGithubAppKey()` replaces the old working key, then the in-lock sign-in validation returns 409. The UI reports that saving failed even though the bot credential was changed and may now be broken. Perform the remaining validation under the mutation lock before replacing the key, and make key/config persistence transactional or restore the prior key if config persistence fails.

packages/core/opensession-server/src/server/github-app.ts
```diff
@@ -191,6 +177,20 @@ export function githubAppConfigured(): boolean {
...
 
+/** Persist the App's private key (0600) at KEY_PATH — the piece the device-flow
+ *  setup never captured, so installation tokens (bot/agent, checks-read) could
+ *  never mint. The App-manifest flow returns this PEM at creation. Drops any
+ *  cached installation token, which belonged to a previous key. Honors the
+ *  OPENSESSION_GITHUB_APP_KEY override so an ops-managed key is never clobbered. */
+export function writeGithubAppKey(pem: string): void {
+  if (process.env.OPENSESSION_GITHUB_APP_KEY)
+    throw new Error("OPENSESSION_GITHUB_APP_KEY is set; not overwriting an ops-managed key");
+  writeFileSync(KEY_PATH, pem.endsWith("\n") ? pem : `${pem}\n`, { mode: 0o600 });
```
tella-butler, line 188: 🟠 **P2** — Key rotation can truncate the existing credential on write failure

`writeFileSync(KEY_PATH, ...)` opens the existing key with truncation. If the filesystem runs out of space or the process dies during rotation, the previous working key can become empty or partial even though the route reports failure. This codebase already uses `writeFileAtomic` for credential and state files; use it here with mode `0o600` so a failed replacement leaves the old key intact.

tella-butler (commented):
OS review · `28c0ea0`

tella-butler:
Superseded by #180, restacked directly on merged #153 after GitHub’s stacked squash left this branch conflicting.

PR #161, opened 2026-08-24
Give the README hero phone room on the right

kentdebruin:
## Summary

- re-shoot the website stage with the phone inset further from the right edge, so it is no longer tight against the frame
- everything else stays the website composition: the window still runs off the right and bottom edges
- new image paths so GitHub cannot serve the cached version

Started by Kent de Bruin in [this OS session](https://os.tella.dev/session/os-01a03308-c365-7836-83a2-afb66aa21dd6)


tella-butler:
### 🤖 OS review

PR #163, opened 2026-08-24
Move creation setup plans into the session actor

jfrolich:
## Summary
- persist branch, workspace, attachment-source, and resolved setup decisions as bounded write-once SessionKernel state
- seed pre-schema-11 actor state once from legacy files, then use the actor for all web and MCP create retries
- atomically retire setup state when the actor commits the exact opening plan and launch effect
- remove the production create-plan file writer while retaining read-only mixed-version import and bounded cleanup
- reject identity crossover, decision overwrite, malformed attachment refs, and secret/body-bearing resolved plans

## Verification
- focused actor/create suite: 109 tests, 991 assertions
- extended create, queue, and restart suite: 151 tests, 1139 assertions
- `bun run typecheck`
- `bun scripts/check-module-side-effects.ts` (412 modules)
- `git diff --check`


tella-butler:
### 🤖 OS review

PR #164, opened 2026-08-24
Defer create replay to durable opening ownership

jfrolich:
## Summary
- stop WebSocket and MCP create-command replay from re-entering setup after the actor has committed a terminal or opening-owned state
- wait for the session projection, or terminal failure, before returning protocol success
- read and identity-fence actor state before invoking setup-plan helpers
- route Web failure through `failCreate` so the pending socket cohort is always retired
- propagate token-scoped recursive handler errors into a failed durable `command_result`

This fixes the live `Creation setup plan was rejected: invalid_transition` recovery path without acknowledging an unresolvable session.

## Verification
- focused ownership and creation suite: 38 tests, 286 assertions
- `bun run typecheck`
- `git diff --check`

### DIFF
--- a/packages/core/opensession-server/src/server/session-create.ts
+++ b/packages/core/opensession-server/src/server/session-create.ts
@@ -407,6 +407,26 @@ const activeOpeningCreates = new Map<
 	}
 >();
 
+export async function waitForCreatedSessionProjection(
+	sessionId: string,
+	identity: string,
+	timeoutMs = 24 * 60 * 60_000,
+): Promise<UnifiedSession> {
+	const deadline = Date.now() + timeoutMs;
+	while (true) {
+		const state = sessionKernel(sessionId).creationState();
+		if (!state || state.identity !== identity)
+			throw new Error("Create request identity crossed durable session ownership");
+		const projected = findSession(sessionId);
+		if (projected) return projected;
+		if (state.state === "failed")
+			throw new Error("Session creation failed before it was persisted");
+		if (Date.now() >= deadline)
+			throw new Error("Session creation remains durably pending");
+		await Bun.sleep(25);
+	}
+}
+
 export function actorCreationSetupPlan(
 	sessionId: string,
 	identity: string,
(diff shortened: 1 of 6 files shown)


tella-butler:
<details>

Safe once the P1 below is fixed. The replay guards correctly avoid re-entering retired setup state, but they treat a nonterminal opening dispatch as a completed creation before the session has been persisted, which can acknowledge and discard the user's create input even when creation subsequently fails.
_1 inline comment below._
<sub>Reviewed `ccf8baf` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-164-review)</sub>

</details>

tella-butler (commented):
OS review · `ccf8baf`

packages/core/opensession-server/src/server/session-create.ts
```diff
@@ -1485,7 +1485,28 @@ export async function handleCreateSessionMessage(
...
 	const createIdentity = requestId || clientSessionId || bksId;
 	let createPlan = actorCreationSetupPlan(bksId, createIdentity);
+	const durableCreation = sessionKernel(bksId).creationState();
 	const recoveringSession = findSession(bksId);
+	if (durableCreation?.state === "failed") {
+		failCreate("Session creation has already failed");
+		return;
+	}
+	if (
+		durableCreation?.state === "opening_dispatched" ||
```
tella-butler, line 1514: 🔴 **P1** — Do not report an unpersisted opening as successfully created

`opening_dispatched` only means the durable outbox owns the opening; `openCreatedSession` persists the session afterward and explicitly handles failures before `announce`. After a restart in that window, this branch sends `session_created` immediately. The frontend then calls `clearDraft` and `dropStagingAttachments` and navigates to the session. If the subsequent `persist()` fails, the actor becomes `failed` with no session file, but the replay has already been acknowledged and the user's draft and staged attachments are gone. The sibling early return in `session-control-wiring.ts` has the same premature-success semantics. Preserve the no-reentry behavior, but for `opening_dispatched` wait for either the session record to exist or the actor to become `failed`; only then return success or surface the terminal failure. `ready` can still return immediately.

tella-butler:
### 🤖 OS review

PR #165, opened 2026-08-24
Claim queued turn batches inside the session actor

jfrolich:
## Summary
- move solo, hold, auto-continue, review-handoff, and delegated-report batch selection into a pure actor reducer
- select the batch and install its pre-journal dispatch in one SQLite transaction
- preserve stable single-message prompt identities and validate externally supplied claim fences
- return only the claimed batch/revision while actor projection invalidation refreshes the gateway cache

## Dependency
- Stacked on #164. Retarget to `main` after that hotfix merges.

## Verification
- focused queue, restart, actor, and ownership suite: 125 tests, 1028 assertions
- `bun run typecheck`
- module side-effect scan (412 modules)
- `git diff --check`

### DIFF
--- a/packages/core/opensession-server/src/server/queue-state.ts
+++ b/packages/core/opensession-server/src/server/queue-state.ts
@@ -500,6 +500,33 @@ export function restorePersistedQueueState(options: {
 	};
 }
 
+/** Let the actor select and claim the next queue batch in one transaction. */
+export function beginNextPromptDispatch(
+	sessionId: string,
+	opts: {
+		soloId?: string;
+		interruptMark?: boolean;
+		stillWorking?: boolean;
+	},
+):
+	| { kind: "empty" }
+	| { kind: "hold"; heldCount: number }
+	| { kind: "deliver"; promptEntryId: string; batch: QueueItem[] } {
+	const claimed = sessionDelivery({
+		op: "claim_next_dispatch",
+		sessionId,
+		promptEntryId: crypto.randomUUID(),
+		...opts,
+	});
+	if (claimed.kind !== "deliver") return claimed;
+	persistQueues();
+	return {
+		kind: "deliver",
+		promptEntryId: claimed.promptEntryId,
+		batch: claimed.items as QueueItem[],
+	};
+}
+
 /** Move a selected queue batch into durable dispatching state before starting
  * runner work. The caller has already removed the batch from promptQueues, so
  * this single persistence point records either the old queued copy (if we die
--- a/docs/session-kernel-architecture.md
+++ b/docs/session-kernel-architecture.md
@@ -90,8 +90,12 @@ actor is unavailable. The direct store adapter exists only for isolated tests.
 Delivery mutation and dispatch claim, acknowledgement or failure are short typed
 Worker reductions. Mutation replies contain only the operation result and new
 revision. They invalidate the gateway projection instead of returning or eagerly
-refetching the full attachment-bearing aggregate. Claiming a batch removes it from the queue and installs its
-dispatch in one SQLite transaction. Failure atomically restores that exact batch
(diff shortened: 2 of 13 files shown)


tella-butler:
<details>

Safe once the actor-less delivery path handles the new claim operation. The production actor atomically selects and claims queue batches as intended, but the compatibility store used by isolated tests returns the wrong result and crashes queue draining.
_1 inline comment below._
<sub>Reviewed `3fd6e36` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-165-review)</sub>

</details>

packages/core/opensession-server/src/server/queue-state.ts
```diff
@@ -500,6 +500,33 @@ export function restorePersistedQueueState(options: {
...
+		soloId?: string;
+		interruptMark?: boolean;
+		stillWorking?: boolean;
+	},
+):
+	| { kind: "empty" }
+	| { kind: "hold"; heldCount: number }
+	| { kind: "deliver"; promptEntryId: string; batch: QueueItem[] } {
+	const claimed = sessionDelivery({
+		op: "claim_next_dispatch",
```
tella-butler, line 516: 🔴 **P1** — Route the new claim operation through the actor-less store

`sessionDelivery()` handles `claim_next_dispatch` only through an installed actor. With no actor, its compatibility dispatch in `session-kernel/kernel.ts` falls through to `failDeliveryDispatch()`, returning a boolean instead of this claim result. This is a verified path: `zz-fake-run.test.ts` intentionally runs directly without installing an actor and exercises queued draining. Once its second prompt reaches this line, `beginNextPromptDispatch()` returns `false`; `drainQueueInner()` then gets an undefined `batch` and fails at `batch.map(...)`. Add a `claim_next_dispatch` branch in the compatibility dispatch that calls `store.claimNextDeliveryDispatch(request)`, matching the new actor-worker branch.

tella-butler (commented):
OS review · `3fd6e36`

tella-butler:
Safe once the P1 below is fixed. The actor path now atomically selects and claims queued batches, but the existing actor-less test-store adapter still does not route the new operation, breaking queue-drain coverage that uses that supported adapter.
_1 inline comment below._
<sub>Reviewed `8dcb5e7` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-165-review)</sub>

packages/core/opensession-server/src/server/queue-state.ts
```diff
@@ -500,6 +500,33 @@ export function restorePersistedQueueState(options: {
...
+	opts: {
+		soloId?: string;
+		interruptMark?: boolean;
+		stillWorking?: boolean;
+	},
+):
+	| { kind: "empty" }
+	| { kind: "hold"; heldCount: number }
+	| { kind: "deliver"; promptEntryId: string; batch: QueueItem[] } {
+	const claimed = sessionDelivery({
```
tella-butler, line 515: 🔴 **P1** — Route the new claim operation through the actor-less store

This still falls through incorrectly when tests install a `SessionKernelStore` without an actor. In `session-kernel/kernel.ts`, the compatibility dispatch handles `claim_dispatch` and then sends every remaining operation to `failDeliveryDispatch`; there is no `claim_next_dispatch` branch. Therefore the existing `zz-fake-run.test.ts` queue-drain scenario calls this function, receives `false` rather than a claim result, and then reaches `batch.map(...)` with `batch === undefined`, while the queued message remains untouched. Add a compatibility branch that calls `store.claimNextDeliveryDispatch(request)` before `claim_dispatch`, and cover the facade path rather than only invoking the store method directly.

tella-butler (commented):
OS review · `8dcb5e7`

PR #168, opened 2026-08-24
Preserve multi-message dispatch identity on retry

jfrolich:
## Summary
- preserve one accepted dispatch ID across every item when a pre-journal batch is restored
- drain restored groups atomically before solo/hold/later queue policy can split or reorder them
- reuse that restored identity for multi-message batches, not only single-message turns
- cover both explicit failure settlement and cold crash restoration through the gateway helper

This prevents early transcript user rows from being duplicated when a multi-message batch retries before a run journal adopts it.

## Verification
- focused queue, actor and restart suite: 86 tests, 757 assertions
- `bun run typecheck`
- `git diff --check`

### DIFF
--- a/packages/core/opensession-server/src/server/session-kernel/store.ts
+++ b/packages/core/opensession-server/src/server/session-kernel/store.ts
@@ -1908,16 +1908,27 @@ export class SessionKernelStore {
         if (state.dispatch) throw new Error("A prompt dispatch is already active");
         const queued = state.queued as QueueItem[];
         if (!queued.length) return { kind: "empty" as const };
-        const plan = selectQueueBatch(queued, {
-          soloId: input.soloId,
-          interruptMark: input.interruptMark,
-          stillWorking: input.stillWorking,
-        });
+        const retryDispatchId = queued.find(
+          (item) => item.retryDispatchId,
+        )?.retryDispatchId;
+        const plan = retryDispatchId
+          ? {
+              kind: "deliver" as const,
+              batch: queued.filter(
+                (item) => item.retryDispatchId === retryDispatchId,
+              ),
+              rest: queued.filter(
+                (item) => item.retryDispatchId !== retryDispatchId,
+              ),
+            }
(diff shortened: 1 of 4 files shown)


tella-butler:
Safe once the P1 below is fixed. The PR preserves retry identity for an unchanged multi-message queue, but normal queue reordering can still discard that identity and duplicate the accepted transcript row.
_1 inline comment below._
<sub>Reviewed `393ef43` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-168-review)</sub>

packages/core/opensession-server/src/server/session-kernel/store.ts
```diff
@@ -1915,9 +1915,7 @@ export class SessionKernelStore {
         });
         if (plan.kind === "hold") return plan;
         const promptEntryId =
-          plan.batch.length === 1 && plan.batch[0]?.promptEntryId
-            ? plan.batch[0].promptEntryId
-            : input.promptEntryId;
+          plan.batch[0]?.promptEntryId || input.promptEntryId;
```
tella-butler: 🔴 **P1** — Reordering a restored batch loses its retry identity

`failDeliveryDispatch` stores the accepted identity only on the original first item, while `reorderQueuedPrompt` and the shipped drag UI allow that item to move. For example, after `[one, two]` fails under `stable-batch-entry`, the queue becomes `[one(promptEntryId=stable-batch-entry), two]`. If the user drags `two` ahead of `one`, the next idle selection still delivers both items, but this line examines only `two` and generates a new identity. `storeAppendUserLineEarly` then inserts a second transcript row instead of upserting `stable-batch-entry`, which is the duplication this PR intends to prevent. Preserve retry-batch ownership independently of queue position, or at minimum select the unique restored identity from the entire selected batch and define how queue removal/splitting transfers that ownership.

tella-butler (commented):
OS review · `393ef43`

PR #169, opened 2026-08-24
Fence background intake during graceful shutdown

jfrolich:
## Summary
- park direct, scheduled, event, and webhook automations once graceful shutdown begins
- reject new automation webhooks with 503 during restart
- stop new GitHub reviews before lock/session creation and cancel a pre-model review that crosses the shutdown boundary
- prevent queued boot-recovery tasks from launching after the shutdown snapshot

This addresses a live restart where the draining process grew from two to four in-flight runs by launching a cron automation, two recovered creates, and a PR review after SIGTERM, forcing a manual cgroup kill.

## Dependency
- Stacked on #168.

## Verification
- shutdown and restart suite: 39 tests, 131 assertions
- `bun run typecheck`
- `git diff --check`

### DIFF
--- a/packages/core/opensession-server/src/server/automations.ts
+++ b/packages/core/opensession-server/src/server/automations.ts
@@ -13,6 +13,7 @@ import {
   webhookBodyTooLargeResponse,
 } from "./shared/bounded-body";
 import { labelIdentity } from "./shared/user-mappings";
+import { isShuttingDown } from "./shutdown-state";
 import { parseCron, cronMatches, nextRun } from "./cron";
 import {
   STRIPE_CONFIRM_TOOLS,
@@ -1122,6 +1123,10 @@ export async function runAutomation(
     modelOverride?: string;
   }
 ): Promise<void> {
+  if (isShuttingDown()) {
+    console.log(`[automations] "${automation.name}" parked during shutdown`);
+    return;
+  }
   const trigger = options?.trigger || "manual";
   // Cron/manual runs don't stack; event/webhook runs are per-event, so they may overlap
   const concurrent = trigger === "event" || trigger === "webhook";
@@ -1676,6 +1681,7 @@ export function startScheduler(onSessionCreated?: (sessionId: string) => void): 
   if (schedulerInterval) return;
 
   schedulerInterval = setInterval(() => {
+    if (isShuttingDown()) return;
     const now = new Date();
     const minuteKey = now.toISOString().slice(0, 16);
     if (minuteKey === lastFiredMinute) return;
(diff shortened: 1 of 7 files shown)


tella-butler:
Safe once the two P1 automation-fencing gaps below are fixed. The recovery and GitHub review fences are correctly positioned, but automation work can still cross the shutdown boundary, while work rejected during shutdown is reported as successfully accepted and can be permanently deduplicated.
_2 inline comments below._
<sub>Reviewed `d534e2b` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-169-review)</sub>

packages/core/opensession-server/src/server/automations.ts
```diff
@@ -1122,6 +1123,10 @@ export async function runAutomation(
     modelOverride?: string;
   }
 ): Promise<void> {
+  if (isShuttingDown()) {
```
tella-butler, line 1126: 🔴 **P1** — Recheck shutdown immediately before launching the engine

This is only checked before the function's first `await`. A code automation can then wait in `listWorktrees`/`createWorktree`, or an ask automation in `ensureAskCheckout` and memory retrieval. If SIGTERM arrives during that work, `snapshotActiveSessions()` runs before this automation has created an agent run, but the continuation still reaches `runAgentHosted`/`runAgent` and launches new work after shutdown began. That recreates the intake growth this PR is intended to prevent. Add a final `isShuttingDown()` check immediately before engine registration/dispatch, and unwind the already-created ledger entry/resources as parked rather than launching or leaving the run marked `running`.

packages/core/opensession-server/src/server/automations.ts
```diff
@@ -1122,6 +1123,10 @@ export async function runAutomation(
...
+    console.log(`[automations] "${automation.name}" parked during shutdown`);
+    return;
```
tella-butler, line 1128: 🔴 **P1** — Do not report parked automation events as successfully handled

Returning normally makes every existing caller treat the trigger as accepted even though no durable work was parked. A concrete failure is `agents/grafana-poller/index.ts`: `investigate()` writes its dedup record before calling `runAutomation`, and rolls it back only in `.catch()`. During shutdown this branch resolves successfully, so no investigation runs and the alert remains muted for the configured dedup period. Slack channel watches similarly mark the event processed before firing. Make the parked outcome observable, such as an explicit result or a handled typed error, and have claim-owning callers roll back or leave their event retryable; HTTP webhook callers should translate it to 503.

tella-butler (commented):
OS review · `d534e2b`

jfrolich:
Superseded by an unstacked branch so the production shutdown fence can ship independently.

PR #170, opened 2026-08-24
Fence background intake during graceful shutdown

jfrolich:
## Summary
- park direct, scheduled, event, Slack, retrigger, and webhook automations once graceful shutdown begins
- persist every accepted automation as a stable pre-launch intent and replay it on boot
- count accepted setup through physical stream adoption in the bounded drain
- reject new automation webhooks with 503 during restart
- stop new GitHub reviews before lock/session creation and preserve recoverable state when an existing review crosses the shutdown boundary
- prevent queued boot-recovery tasks from launching after the shutdown snapshot

This addresses a live restart where the draining process grew from two to four in-flight runs by launching a cron automation, two recovered creates, and a PR review after SIGTERM, forcing a manual cgroup kill.

## Verification
- shutdown and restart suite: 40 tests, 152 assertions
- `bun run typecheck`
- `git diff --check`

### DIFF
--- a/packages/core/opensession-server/src/server/automations.ts
+++ b/packages/core/opensession-server/src/server/automations.ts
@@ -6,6 +6,7 @@
 import { randomUUIDv7 } from "bun";
 import { OPENSESSION_SESSIONS_DIR , newSessionId} from "./paths";
 import { mkdirSync, readdirSync, readFileSync, unlinkSync, existsSync } from "fs";
+import { join } from "path";
 import { writeJsonAtomic } from "./shared/atomic-write";
 import {
   RequestBodyTooLargeError,
@@ -13,6 +14,7 @@ import {
   webhookBodyTooLargeResponse,
 } from "./shared/bounded-body";
 import { labelIdentity } from "./shared/user-mappings";
+import { isShuttingDown } from "./shutdown-state";
 import { parseCron, cronMatches, nextRun } from "./cron";
 import {
   STRIPE_CONFIRM_TOOLS,
@@ -22,6 +24,7 @@ import {
 import { getAccountById } from "./claude-accounts";
 import { getCodexAccountById } from "./codex-accounts";
 import { runAgent } from "./agent-runner";
+import { activeRunRecords } from "./run-journal";
 import { runAgentHosted } from "./host-client";
 import {
   providerFor,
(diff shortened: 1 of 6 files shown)


tella-butler:
<details>

### 🤖 OS review

🔄 Reviewing `a3f7896`… · [📺 open session](https://os.tella.dev/session/bks-ghpr-opensession-170-review)

</details>

tella-butler:
<details>

### 🤖 OS review

</details>

tella-butler:
Safe once the two P1 recovery issues below are fixed. The shutdown fences cover the intended intake paths, but pending automation intents can enter a tight retry loop or replay an already-completed agent run across restart.
_2 inline comments below._
<sub>Reviewed `5ed5e06` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-170-review)</sub>

packages/core/opensession-server/src/server/automations.ts
```diff
@@ -1098,10 +1108,104 @@ function sandboxAutomationMcpEgress(mcpServers: string[]): string[] {
...
+      ) continue;
+      void runAutomation(automation, onSessionCreated, {
+        trigger: intent.trigger,
+        eventContext: intent.eventContext,
+        modelOverride: intent.modelOverride,
+        osSessionId: intent.sessionId,
+        acceptedAt: intent.acceptedAt,
+        deleteAutomationAfterRun: intent.deleteAutomationAfterRun,
+      }).finally(() => {
+        if (!isShuttingDown()) resumePendingAutomationRuns(onSessionCreated);
```
tella-butler, line 1235: 🔴 **P1** — Failed pending intents retry without delay forever

`runAutomation` catches pre-launch failures and deliberately leaves the intent file when `physicalAdopted` is false. This unconditional `.finally()` then immediately rescans and launches that same intent again. For example, a persisted sandbox automation with the MicroVM provider unavailable fails `validateSandboxAutomation`, settles its ledger entry, and is restarted in the next microtask indefinitely. That creates a CPU/log/filesystem-write loop and can starve normal server work after boot. Remove the immediate recursive rescan, or retry retained intents through a bounded, delayed scheduler with backoff.

packages/core/opensession-server/src/server/automations.ts
```diff
@@ -1548,11 +1681,14 @@ export async function runAutomation(
       preparedInputs.commit();
     }
 
+    const completedIntent = clearAutomationIntent(bksId);
```
tella-butler: 🔴 **P1** — Intent remains replayable after the engine journal has settled

The intent is only cleared after session persistence and `deliverAutomationOutputs`, even though `pi-runner` clears the run journal as soon as the engine reaches a terminal event. If the process exits after the stream finishes but while an output is being delivered, the next boot sees the intent without a matching `activeRunRecord` and reruns the entire automation under the same session ID. Any tools or external mutations performed by the completed agent can therefore happen twice. Transfer recovery ownership at physical stream adoption by retiring the pre-launch intent once the journal owns the run; persist `deleteAutomationAfterRun` in the run/session settlement state so recovered one-offs can still be deleted without keeping the whole automation replayable.

tella-butler (commented):
OS review · `5ed5e06`

PR #171, opened 2026-08-24
Fix create session mailbox deadlock

jfrolich:
## Summary
- keep WebSocket session creation outside the legacy per-session mailbox
- remove redundant create-session mailbox wrappers from REST and session control
- rely on the dedicated durable creation FSM for idempotency and recovery

## Root cause
The outer create command waited for the session projection while the opening effect waited for an inner `session_file_updated` command on the same mailbox. Both commands remained `processing`, leaving the UI on “Setting up workspace”.

## Verification
- `bun test packages/core/opensession-server/src/server/session-kernel/ownership.test.ts`
- focused creation/kernel actor tests
- `bun run typecheck`

### DIFF
--- a/packages/core/opensession-server/src/server/routes/sessions.ts
+++ b/packages/core/opensession-server/src/server/routes/sessions.ts
@@ -73,7 +73,6 @@ import {
 	tombstoneSessionKernel,
 } from "../session-kernel";
 import {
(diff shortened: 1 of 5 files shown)


tella-butler:
### 🤖 OS review

PR #175, opened 2026-08-24
Make interrupt delivery actor-owned

jfrolich:
## Summary
- atomically prepare a stable interrupt identity plus fenced cancel outbox effect before physical cancellation
- retry an ambiguous executing cancel only against the same actor run generation and settle explicit `confirmed` or `not_aborted` outcomes
- wait for confirmation, then transfer the interrupt only with the queue batch containing its durable anchor
- restore interrupt ownership with unjournaled dispatches during both ordinary failure and production boot recovery
- advance the actor and SessionKernel schema protocols to 12

## Verification
- `bun run typecheck`
- 127 focused queue, actor, kernel, effect, and ownership tests
- crash coverage before cancel, during cancel, after claim, boot recovery, retry-group crossover, failed dispatch, steered receipt rollback, and non-destructive boot recovery
- `bun scripts/check-module-side-effects.ts` (413 modules)
- `git diff --check`


tella-butler:
<details>

```mermaid
sequenceDiagram
  participant Gateway
  participant Actor
  participant SQLite
  Gateway->>Actor: mark_interrupt(anchorId, soloId, expiresAt)
  Actor->>SQLite: Persist interrupt
  Gateway->>Actor: claim_next_dispatch
  Actor->>SQLite: Consume interrupt, select batch, install dispatch
  SQLite-->>Actor: Committed delivery state
  Actor-->>Gateway: batch + interrupted
```

</details>

tella-butler:
### 🤖 OS review

PR #184, opened 2026-08-24
Retire settled cancelled abnormal journals

jfrolich:
## Summary
- retire an abnormal-completion journal when its exact durable turn cancel already settled
- keep detached-host private journals and actor-uncertain evidence unchanged
- prevent cancelled runs that ended without a terminal frame from remaining permanently busy until restart

## Live evidence
A schema-13 Stop canary settled `confirmed` and physically aborted its bash command, but its source ended without a terminal event. The retained `terminalFailure` journal kept the session visibly running despite zero pending/dead-lettered outbox work. This closes that post-settlement cleanup gap.

## Verification
- `bun run typecheck`
- `bun test packages/core/opensession-server/src/server/zz-run-journal.test.ts` (45 pass)
- `git diff --check`


tella-butler:
### 🤖 OS review

PR #186, opened 2026-08-24
Fence cancel settlement cleanup ownership

jfrolich:
## Summary
- distinguish an exact already-settled turn cancel from missing or replaced ownership
- skip journal cleanup for stale outbox effects and honor the boolean result of settlement after physical cancellation
- cover exact settled, replacement, missing, and rejected settlement behavior

## Verification
- `bun run typecheck`
- 85 focused kernel and ownership tests
- `git diff --check`


tella-butler:
### 🤖 OS review

PR #187, opened 2026-08-24
Project terminal turn outcomes through actor effects

jfrolich:
## Summary
- add schema 14 actor receipts and a typed `turn_outcome_project` effect for normal and opening terminal outcomes
- fence projection admission by immutable run id and generation, retain permanent per-generation receipts, and defer later generations behind earlier live projections without consuming retry budget
- apply transcript notices, `lastRunError`, and worker-failure notification through a strict destination-idempotent executor before exact actor settlement and outbox acknowledgement
- preserve compatibility-only direct projections in tests and detached private hosts while failing closed without the production actor

## Verification
- `bun run typecheck`
- 108 actor, kernel, and ownership tests
- 151 broader actor, run-state, transcript, and fake-run tests
- `git diff --check`


tella-butler:
### 🤖 OS review

PR #188, opened 2026-08-24
Stop deep links landing on the viewer's last session

tella-butler:
A link into OS from Plain ("Open OS ↗") or Slack could open whatever session the viewer last had open, instead of the ticket or session it named.

## Why

Three separate causes, all ending in the same place. `App.tsx:721-748` restores the last session from `localStorage` on a cold load when the path resolves to the home view — added because iOS evicts standalone PWAs and relaunches them at `start_url`. But `{view: "prs"}` is also `parseRoute`s catch-all for **any** path it does not recognise, so "the deep link did not survive" and "go to my last session" are the same branch.

| # | Cause | Fix |
| --- | --- | --- |
| 1 | `/plain-triage/<id>` holds the navigation open 15-120s while it boots the triage automation; the service worker stall guard paints the cached shell after 5s, stranding the document at a URL the router cannot parse | Route answers immediately |
| 2 | The SW applies its shell fallback to every navigation, including server routes that answer with a redirect | Skip those routes |
| 3 | The manifest declares no `launch_handler`, so a link click focuses the installed window and ignores the URL — nothing navigates, and Back does nothing | `client_mode: navigate-existing` |

## What changed

- **`src/server/routes/plain.ts`** — split the cheap cache lookup (`existingPlainTriageSession`) out of the boot-and-wait (`resolvePlainTriageSession`). The `/plain-triage/<id>` redirect now always answers immediately: an existing linked session still redirects to `/session/<id>`; otherwise the triage run is kicked off in the background and the redirect goes to `/support/<threadId>`, which renders the real ticket right away and gains the session as a tab under the ticket workspace once it boots. The JSON twin behind the Support views "Triage this ticket" button still awaits, since its client shows a progress state. Each resolution is now logged.
- **`src/frontend/sw.js`** — added `REDIRECT_RE` and excluded those paths from `shellNavigate`, so a redirecting route can never be replaced by the cached shell at a URL the router has no route for.
- **`src/server/routes/static-assets.ts`** — `launch_handler: { client_mode: "navigate-existing" }` on the PWA manifest.

## Notes

- Backend change: needs `systemctl restart opensession` after merge. The SW and manifest ship with the frontend rebuild; installed PWAs pick up the new manifest on their next update check.
- `sw.test.ts` passes. The other failures in `src/server/routes/` are missing `node_modules` in the fresh worktree, unrelated to this diff.

Started by Johnny Lin in [this OS session](https://os.tella.dev/session/slack-C0BE8KVFBEX-1787586472.612789)

### DIFF
--- a/packages/core/opensession-server/src/frontend/sw.js
+++ b/packages/core/opensession-server/src/frontend/sw.js
@@ -78,6 +78,13 @@ const NAV_STALL_MS = 5000;
 // matches sw.js itself (no dash) or icons/splash (not js/css).
 const ASSET_RE = /^\/(?:opensession\/|backstage\/)?[\w.]+-\w+\.(?:js|css)$/;
 const API_RE = /^\/(?:opensession\/|backstage\/)?api\//;
+// Server routes that answer a navigation with a REDIRECT rather than the app
+// shell. The stall guard below must never fire on these: they can legitimately
+// take longer than NAV_STALL_MS (a Plain triage boot runs 15-120s), and
+// painting the shell over one strands the document at a URL the router has no
+// route for — which the app then treats as "landed on home" and replaces with
+// the viewer's last session. Let them go straight to the network and redirect.
+const REDIRECT_RE = /^\/(?:opensession\/|backstage\/)?plain-triage\//;
 // A build ships ~a dozen chunks; 80 keeps a few builds' worth before pruning.
 const MAX_ASSETS = 80;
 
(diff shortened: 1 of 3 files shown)


tella-butler:
Safe to merge. The redirect now returns a thread-specific destination without waiting for automation startup, while the service worker and PWA manifest preserve incoming navigation URLs. Callers and route consumers remain compatible, and no correctness or safety issues were found.
<sub>Reviewed `2f57f0f` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-188-review)</sub>

PR #189, opened 2026-08-24
Retire settled automation intents during recovery

jfrolich:
## Summary
- retire accepted automation intents when their same-session durable run ledger is already terminal
- prevent restart recovery from replaying physically completed automation work
- cover terminal, running, and cross-session recovery decisions

## Verification
- `bun test packages/core/opensession-server/src/server/automation-intent-recovery.test.ts packages/core/opensession-server/src/server/shutdown-intake.test.ts`
- `bun run typecheck`
- `git diff --check`

### DIFF
--- a/packages/core/opensession-server/src/server/automations.ts
+++ b/packages/core/opensession-server/src/server/automations.ts
@@ -79,6 +79,7 @@ import {
   sanitizeAutomationOutputs,
   type AutomationOutput,
 } from "./automation-outputs";
+import { automationIntentAlreadySettled } from "./automation-intent-recovery";
 
 const AUTOMATIONS_DIR = stateDir("automations");
 const SESSIONS_DIR = OPENSESSION_SESSIONS_DIR;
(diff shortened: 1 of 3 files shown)


tella-butler:
Safe once the P1 below is fixed. The new recovery check closes the intended settle-before-clear crash window, but it also treats an ambiguous error ledger entry as proof that a still-journaled physical run completed, which can prematurely retire one-off automation state.
_1 inline comment below._
<sub>Reviewed `be20f4e` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-189-review)</sub>

packages/core/opensession-server/src/server/automations.ts
```diff
@@ -1207,6 +1208,13 @@ export function resumePendingAutomationRuns(
       ) throw new Error("invalid automation intent");
       const automation = getAutomation(intent.automationId);
       if (!automation) throw new Error(`automation ${intent.automationId} is unavailable`);
+      if (automationIntentAlreadySettled(intent.sessionId, automation.runs || [])) {
```
tella-butler, line 1211: 🔴 **P1** — A terminal ledger entry does not prove an active physical run completed

`runAutomation` catches exceptions after physical adoption, marks the ledger `error`, and intentionally retains the intent because the active journal may still own execution (lines 1747-1755). On restart, this branch runs before the existing `activeRunRecords()` ownership check and clears that intent anyway. For example, if persisting session metadata throws while a sandbox run remains journaled, a one-off intent reaches this branch with `status: "error"`; recovery deletes the automation and its input/output state, then the later journal recovery cannot rebuild its automation MCP servers because `automationResumeMcpForSession` requires that record. Preserve the distinction between ledger settlement and physical completion: do not mark an active-journal catch path terminal, or require stronger terminal evidence before retiring its intent.

tella-butler (commented):
OS review · `be20f4e`

PR #191, opened 2026-08-24
Make Stop terminal for opening creation effects

jfrolich:
## Summary
- advance schema and actor protocol to 15 with a terminal `cancelled` creation state
- settle an opening creation effect when Stop is prepared, before launch, during local recovery, or at exact terminal/recovery settlement
- retain exact physical run generation for opening outcome projection and fence late success after cancellation
- clear the stable prompt dispatch and recovery plan only after the actor records the cancellation receipt

## Verification
- 173 focused creation, kernel, recovery, ownership, and journal tests pass
- `bun run typecheck`
- `git diff --check`


tella-butler:
<details>

Safe once the P1 below is fixed. The creation actor now records Stop as terminal and fences late success, but live Runner-backed opening turns are not physically connected to the cancellation token, so they can continue executing after Stop.
_1 inline comment below._
<sub>Reviewed `8c94abe` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-191-review)</sub>

</details>

packages/core/opensession-server/src/server/session-create.ts
```diff
@@ -1165,18 +1214,26 @@ export async function openCreatedSession(
...
 				// This runs in the terminal event's consumer body, before requesting
 				// the generator's next item. Backend generator finally blocks may now
 				// retire their journal/host only after the actor has the receipt.
-				settleCreationSucceeded(
-					bksId,
-					creationIdentity,
-					undefined,
-					creationEffectId,
-				);
+				if (openingTurnWasCancelled())
```
tella-butler, line 1223: 🔴 **P1** — Stop does not cancel a live Runner-backed opening turn

This terminal check protects only the creation receipt. For a Runner-backed create, `markSessionStarting` admits `startToken`, while `maybeLaunchRunnerRun` is launched under the distinct `runnerOpeningHostId(openingRun.runId, openingRun.generation)` and receives no `shouldCancel` callback. `requestTurnCancel` therefore calls `cancelAgentRunTokenAndWait(startToken)`, but `hostCancel` cannot reach the Runner host registered under the derived ID; it waits for the pending start token to disappear while the remote agent continues running. For example, pressing Stop after a Runner opening starts can leave its tools, commits, and billing running until natural completion, even though creation is already durably `cancelled`. Make the admitted run ID and Runner host ID identical, as the normal Runner prompt path does, or otherwise durably map the admitted token to the physical Runner control and propagate cancellation throughout the launch/stream lifetime.

tella-butler (commented):
OS review · `8c94abe`

packages/core/opensession-server/src/server/session-create.ts
jfrolich: Fixed in ed9808ab1: opening creation now admits the actor under the same stable token passed to `maybeLaunchRunnerRun` as `hostId`, so exact Stop cancellation reaches the live Runner host. A local duplicate guard prevents two concurrent launchers from sharing that token while restart adoption can reuse it after gateway projection loss. Added focused stable-token and ownership tests.

tella-butler:
Safe once the P1 below is fixed. The follow-up correctly gives actor admission and the live Runner host one stable token, resolving the previous live-cancellation issue, but Runner creation still ignores that token's pre-launch cancellation latch.
_1 inline comment below._
<sub>Reviewed `ed9808a` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-191-review)</sub>

packages/core/opensession-server/src/server/session-create.ts
```diff
@@ -1114,14 +1168,7 @@ export async function openCreatedSession(
...
 						promptEntryId: openingPromptEntryId,
-						...(openingRun
-							? {
-								hostId: runnerOpeningHostId(
-									openingRun.runId,
-									openingRun.generation,
-								),
-							}
-							: {}),
+						hostId: startToken,
```
tella-butler, line 1171: 🔴 **P1** — Runner creation can still launch after Stop

`maybeLaunchRunnerRun` only rejects a prepared launch when its `shouldCancel` callback returns true (`runner-session.ts:217-220`). The normal prompt path supplies that callback, but this creation path supplies only `hostId`. If Stop arrives while `prepareRunnerWorkspace` is running, `cancelAgentRunToken` latches `startToken`, yet this call still sends `launchRunnerHost`; cancellation is only retried after the host becomes controllable, allowing a stopped opening prompt to start and potentially execute tools. Pass `shouldCancel: () => isAgentSessionCancelled(bksId, startToken)` here, importing the existing helper as the normal Runner path does.

tella-butler (commented):
OS review · `ed9808a`

packages/core/opensession-server/src/server/session-create.ts
jfrolich: Completed in d8f10586a: the Runner opening now also receives `shouldCancel: () => isAgentSessionCancelled(bksId, startToken)`, so a Stop latched during workspace preparation prevents physical Runner launch rather than relying only on the exact host control becoming available afterward. Focused tests, typecheck, and side-effect scan remain clean.

packages/core/opensession-server/src/server/session-create.ts
jfrolich: Fixed in d8f10586a: passed the exact requested `shouldCancel: () => isAgentSessionCancelled(bksId, startToken)` callback to the Runner opening launcher, so a Stop latched during workspace preparation prevents launch. The same stable token still owns actor admission and live host control.

PR #192, opened 2026-08-24
Keep Stop bookkeeping after a racing creation settlement

jfrolich:
Follow-up to #191 (merged `541b8e1f9`).

## Problem

`requestTurnCancel()` called `settleCreationOpeningForStop()` unguarded after the durable `prepare_cancel` commit. Each kernel round-trip can interleave other actor work, so an opening effect settling `failed`/`succeeded` between the cancel commit and the creation reducer write makes the settlement throw (`stale_effect`, `invalid_transition`, or identity mismatch). That throw skipped `stoppedSessions.add`, `persistQueues()`, and `broadcastQueue()`, so a Stop that durably committed surfaced as an error and its requeued prompts were not persisted or broadcast.

## Fix

- Wrap the opening settlement in try/catch inside `requestTurnCancel()`: the durable Stop always completes its local bookkeeping; a racing settlement is logged and skipped.
- Tighten `settleStoppedCreationOpening()` to exact cancel identity: a retained receipt fences only when `cancel.runId`/`cancel.runGeneration` match the effect payload, matching `openingTurnWasCancelled()` and `settleRecoveredCreationOpening()`.
- Extend ownership ordering assertions for both fences; update the architecture doc.

## Verification

- 174 focused kernel/ownership/creation/journal tests pass.
- `bun run typecheck`, 415-module side-effect scan, and `git diff --check` clean.

### DIFF
--- a/docs/session-kernel-architecture.md
+++ b/docs/session-kernel-architecture.md
@@ -219,7 +219,11 @@ durable stopped turn or its retained cancel receipt before launch and while
 awaiting a detached local owner, so a restart cannot resurrect a cancelled
 opening prompt. Runner, sandbox, and local openings use the same stable token
 for actor admission and physical control, letting Stop reach the exact backend
-without giving up restart adoption.
+without giving up restart adoption. Stop bookkeeping never depends on the
+creation settlement succeeding: a concurrent opening result racing the cancel
+read is logged and skipped, while the durable `prepare_cancel` commit, queue
+persistence, and broadcast always run. Retained cancel receipts fence by exact
+run id and generation, matching the opening they cancelled.
 Non-image create attachments are durably spooled to bounded source references,
 then copied or adopted at deterministic session-owned paths by
 `creation_attachment_stage`; digest crossover fails closed and inline bodies
--- a/packages/core/opensession-server/src/server/run-session.ts
+++ b/packages/core/opensession-server/src/server/run-session.ts
@@ -353,17 +353,16 @@ export function requestTurnCancel(
     ...(request.user ? { user: request.user } : {}),
   });
   // The durable Stop committed above must not be undone by a concurrent
-  // creation settlement racing this read (the opening effect may have just
-  // settled failed/succeeded between our snapshot and the reducer write).
-  // Stop bookkeeping below therefore always runs.
+  // creation settlement racing this read: terminal settlements are idempotent
+  // (see settleCreationCancelled), so only genuine invariant failures throw.
+  // Bookkeeping below always runs, even when the error propagates.
   try {
     settleCreationOpeningForStop(sessionId);
-  } catch (e) {
-    console.error(`[stop] Opening creation settlement failed for ${sessionId}:`, e);
+  } finally {
+    stoppedSessions.add(sessionId);
+    persistQueues();
+    broadcastQueue(sessionId);
   }
-  stoppedSessions.add(sessionId);
-  persistQueues();
-  broadcastQueue(sessionId);
   return { requeued: requeued.length };
 }
 
--- a/packages/core/opensession-server/src/server/run-session.ts
+++ b/packages/core/opensession-server/src/server/run-session.ts
@@ -352,7 +352,15 @@ export function requestTurnCancel(
     source: request.source,
     ...(request.user ? { user: request.user } : {}),
   });
-  settleCreationOpeningForStop(sessionId);
+  // The durable Stop committed above must not be undone by a concurrent
+  // creation settlement racing this read (the opening effect may have just
+  // settled failed/succeeded between our snapshot and the reducer write).
+  // Stop bookkeeping below therefore always runs.
+  try {
+    settleCreationOpeningForStop(sessionId);
+  } catch (e) {
+    console.error(`[stop] Opening creation settlement failed for ${sessionId}:`, e);
+  }
   stoppedSessions.add(sessionId);
   persistQueues();
   broadcastQueue(sessionId);
--- a/packages/core/opensession-server/src/server/session-create.ts
+++ b/packages/core/opensession-server/src/server/session-create.ts
@@ -659,7 +659,16 @@ async function restorePlannedOpening(sessionId: string): Promise<{
 export function settleStoppedCreationOpening(item: CreationOpeningEffectItem): boolean {
 	const kernel = sessionKernel(item.sessionId);
 	const turn = sessionTurn({ op: "snapshot", sessionId: item.sessionId });
-	if (kernel.runState().state !== "stopped" && !turn.cancel) return false;
+	// Exact cancel identity, matching openingTurnWasCancelled(): a retained
+	// receipt fences only the exact run it cancelled. `stopped` alone still
+	// fences because creation is single-shot — while it is opening_dispatched
+	// the opening turn is the only run Stop could have targeted.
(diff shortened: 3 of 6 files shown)


tella-butler:
<details>

Safe once the P1 below is fixed. The bookkeeping guard achieves the primary race fix, but the new exact-cancel fence compares against a different run ID representation than the opening turn actually registers, allowing a stopped opening to be revived after the stop latch is lifted.
_1 inline comment below._
<sub>Reviewed `aa9f6b2` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-192-review)</sub>

</details>

packages/core/opensession-server/src/server/run-session.ts
```diff
@@ -352,7 +352,15 @@ export function requestTurnCancel(
...
+  // creation settlement racing this read (the opening effect may have just
+  // settled failed/succeeded between our snapshot and the reducer write).
+  // Stop bookkeeping below therefore always runs.
+  try {
+    settleCreationOpeningForStop(sessionId);
+  } catch (e) {
+    console.error(`[stop] Opening creation settlement failed for ${sessionId}:`, e);
+  }
   stoppedSessions.add(sessionId);
   persistQueues();
```
jfrolich: Addresses the blocking finding from the #191 adversarial review (bks-92603cc3): `settleCreationOpeningForStop` is now guarded after the durable `prepare_cancel` commit, so a racing opening settlement cannot skip Stop bookkeeping. Also tightened `settleStoppedCreationOpening` to exact cancel run identity per the non-blocking observation.

packages/core/opensession-server/src/server/session-create.ts
```diff
@@ -659,7 +659,16 @@ async function restorePlannedOpening(sessionId: string): Promise<{
...
 	const turn = sessionTurn({ op: "snapshot", sessionId: item.sessionId });
-	if (kernel.runState().state !== "stopped" && !turn.cancel) return false;
+	// Exact cancel identity, matching openingTurnWasCancelled(): a retained
+	// receipt fences only the exact run it cancelled. `stopped` alone still
+	// fences because creation is single-shot — while it is opening_dispatched
+	// the opening turn is the only run Stop could have targeted.
+	const cancel = turn.cancel;
+	const exactCancel =
+		cancel != null &&
+		cancel.runId === item.payload.runId &&
```
tella-butler: 🔴 **P1** — Exact cancel check uses the unregistered opening run ID

`requestCreationOpening()` stores a payload ID shaped like `opening:${sessionId}:${promptEntryId}`, but `openCreatedSession()` passes that through `runnerOpeningHostId()` before `markSessionStarting()`. Consequently, Stop records the resulting `rh-opening-…` token in `turn.cancel.runId`, so this comparison can never match the effect payload. A concrete failure occurs when an opening effect is delayed or retried, the user stops its registered `rh-opening-…` run, then sends another prompt: `liftUserStop()` moves the actor out of `stopped`, this function sees neither `stopped` nor `exactCancel`, and the cancelled opening is launched/recovered instead of settled. Compare the receipt with `runnerOpeningHostId(item.payload.runId, item.payload.runGeneration)` (and retain the generation check), matching the token used by `openingTurnWasCancelled()`.

tella-butler (commented):
OS review · `aa9f6b2`

jfrolich:
Second push `3d6419af3`: releases the process run reservation (`unmarkSessionStarting`) before throwing "Opening turn lost actor admission before preparation" — a finding from the #191 adversarial review (bks-54367907): that throw landed outside the inner try/finally and could leave `pendingStarts` wedged busy until restart. Ownership test now asserts the unmark precedes the throw.

tella-butler:
<details>

Safe once the remaining P1 is fixed. The guarded settlement now preserves Stop bookkeeping, and the follow-up releases rejected admission reservations, but it does not resolve the retained-cancel identity mismatch from the prior review.
_1 inline comment below._
<sub>Reviewed `3d6419a` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-192-review)</sub>

</details>

tella-butler:
Safe to merge. The prior P1 is fixed by comparing the cancel receipt with the deterministic physical opening token, and terminal settlement races now preserve Stop bookkeeping without hiding invariant failures. CI passes; only the advisory documentation mismatch below remains.
_1 inline comment below._
<sub>Reviewed `66421c5` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-192-review)</sub>

docs/session-kernel-architecture.md
```diff
@@ -219,7 +219,11 @@ durable stopped turn or its retained cancel receipt before launch and while
 awaiting a detached local owner, so a restart cannot resurrect a cancelled
 opening prompt. Runner, sandbox, and local openings use the same stable token
 for actor admission and physical control, letting Stop reach the exact backend
-without giving up restart adoption.
+without giving up restart adoption. Stop bookkeeping never depends on the
+creation settlement succeeding: a concurrent opening result racing the cancel
+read is logged and skipped, while the durable `prepare_cancel` commit, queue
```
tella-butler, line 224: ⚪ **P3** — Architecture doc describes the removed catch-and-log behavior

A terminal opening race is now handled idempotently by `settleCreationCancelled()`, so it produces no exception or log. Conversely, an identity or other invariant failure propagates from `requestTurnCancel()` after its `finally` bookkeeping and is not logged here. Update this sentence to say terminal races are idempotent while invariant failures remain fail-loud after queue persistence and broadcast.

tella-butler (commented):
OS review · `66421c5`

jfrolich:
Addressed both blockers from the exact-head review in `66421c502`:

1. **Wrong identity plane** — `settleStoppedCreationOpening` now compares the retained receipt against `runnerOpeningHostId(item.payload.runId, item.payload.runGeneration)` (the admitted physical token the cancel receipt actually records) plus generation, instead of the logical payload run id. A Stop followed by `liftUserStop` can no longer let outbox replay launch a cancelled opening.
2. **Silent swallow** — `settleCreationCancelled` is now idempotent on terminal states (`ready`/`failed`, mirroring its sibling settlers), so the proven-benign race no-ops; genuine invariant failures still propagate, and `requestTurnCancel` uses try/finally so bookkeeping always runs while errors stay loud.

New regression test covers cancellation-after-terminal for both ready and failed; ownership assertions updated.

jfrolich:
Addressed both blockers from the exact-head review in `66421c502`:

1. **Wrong identity plane** - `settleStoppedCreationOpening` now compares the retained receipt against `runnerOpeningHostId(item.payload.runId, item.payload.runGeneration)` (the admitted physical token the cancel receipt actually records) plus generation, instead of the logical payload run id. A Stop followed by `liftUserStop` can no longer let outbox replay launch a cancelled opening.
2. **Silent swallow** - `settleCreationCancelled` is now idempotent on terminal states (`ready`/`failed`, mirroring its sibling settlers), so the proven-benign race no-ops; genuine invariant failures still propagate, and `requestTurnCancel` uses try/finally so bookkeeping always runs while errors stay loud.

New regression test covers cancellation-after-terminal for both ready and failed; ownership assertions updated.

PR #193, opened 2026-08-24
Bind sandbox launch identity to the admitted run token

jfrolich:
Same defect class as the Runner opening P1 fixed in #191, found while re-verifying its assumption across backends.

## Problem

`maybeLaunchSandboxedRun` minted a random `hostId` for the physical sandbox host and never registered the admitted run token anywhere the exact-token cancel path can see:

- `spec.hostId` (and therefore the journal `runKey`, host-registry keys, and recovery identity) was random.
- Nothing added `startToken` to `activeSessionRunTokens`, so `agentRunTokenLatched(startToken)` stayed false.
- `hostCancel(startToken)` missed (the control registers under session/host ids).

Consequence: Stop during a live sandbox turn — opening or normal prompt, any provider — could not reach the physical host. `cancelAgentRunTokenAndWait` spun to its 120s timeout and threw, leaving the turn_cancel effect retrying while the remote/in-sandbox agent kept executing tools until natural completion. The local in-process sandbox path (`sandbox/local.ts`) also dropped the token on the floor when delegating to `runAgent`.

## Fix

- `maybeLaunchSandboxedRun` now uses `hostId: opts.startToken || random`, mirroring `spawnHostRun` and the Runner path: journal, registry, and recovery all key on the admitted token.
- `sandbox/local.ts` passes `startToken: spec.hostId` into `runAgent` so in-process latch/cancel works.
- Ownership assertions cover both bindings.

## Verification

- `bun run typecheck`, focused ownership/run-state/journal tests, 415-module side-effect scan, `git diff --check` clean.

### DIFF
--- a/packages/core/opensession-server/src/server/run-session.ts
+++ b/packages/core/opensession-server/src/server/run-session.ts
@@ -1670,7 +1670,10 @@ export async function maybeLaunchSandboxedRun(
 			? portableWorkspacePresetRun(workspacePreset)
 			: undefined;
 		const spec: RunHostSpec = {
-			hostId: `rh-${randomUUIDv7()}`,
+			// Bind the physical sandbox host to the admitted run token, exactly like
+			// the Runner and local paths: exact-token Stop must reach the live host,
+			// and restart adoption must reattach under the same durable identity.
+			hostId: opts.startToken || `rh-${randomUUIDv7()}`,
 			osSessionId: session.id,
 			prompt: opts.prompt,
 			promptEntryId: opts.promptEntryId,
--- a/packages/core/opensession-server/src/server/sandbox/docker.ts
+++ b/packages/core/opensession-server/src/server/sandbox/docker.ts
@@ -1208,6 +1208,10 @@ function makeDockerSandbox(
         record.launchPhase = "started";
         journalSet(record);
         mark("host exec dispatched");
+        if (handle.cancelled)
+          throw new HostLaunchNotDispatchedError(
+            `${spec.hostId} was cancelled while launching`,
+          );
         await handle.connectWithWait(30_000);
         mark("host attached");
       } catch (error) {
--- a/packages/core/opensession-server/src/server/sandbox/docker.ts
+++ b/packages/core/opensession-server/src/server/sandbox/docker.ts
@@ -1196,6 +1196,11 @@ function makeDockerSandbox(
       const mark = (step: string) =>
         console.log(`[sandbox] launch ${spec.hostId.slice(0, 11)}: ${step} (+${Date.now() - t0}ms)`);
       try {
+        // Construct (and register) the control BEFORE dispatch so exact-token
+        // Stop reaches the launching host: cancelAgentRunTokenAndWait sees
+        // hostRunBusy(token) during the launch await, and cancelHost's stop
+        // backstop plus the cancelled startup marker fence the dispatch race.
+        handle = new HostHandle(dir, spec, callbacks, launcher);
         await launcher.launch(spec.hostId, dir, () => {
           record.launchPhase = "launching";
           journalSet(record);
@@ -1203,7 +1208,6 @@ function makeDockerSandbox(
         record.launchPhase = "started";
         journalSet(record);
         mark("host exec dispatched");
-        handle = new HostHandle(dir, spec, callbacks, launcher);
         await handle.connectWithWait(30_000);
         mark("host attached");
       } catch (error) {
--- a/packages/core/opensession-server/src/server/sandbox/docker.ts
+++ b/packages/core/opensession-server/src/server/sandbox/docker.ts
@@ -1217,7 +1217,11 @@ function makeDockerSandbox(
       } catch (error) {
         const definitelyNotDispatched =
           record.launchPhase === "prepared" ||
-          error instanceof HostLaunchNotDispatchedError;
+          error instanceof HostLaunchNotDispatchedError ||
+          // A stop backstop that proved absence during the launch/connect
+          // await already finished this handle: retire it like a
+          // never-dispatched launch instead of reconciling an ended owner.
+          handle?.ended === true;
         if (definitelyNotDispatched) {
           journalClearIfLineage(record);
           handle?.abandon();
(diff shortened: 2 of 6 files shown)


tella-butler:
<details>

Safe once the P1 below is fixed. The initial sandbox launch now correctly binds its host and journal identity to the admitted run token, but recovery relaunches still discard that identity, leaving Stop unable to cancel the replacement host.
_1 inline comment below._
<sub>Reviewed `64ebde8` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-193-review)</sub>

</details>

packages/core/opensession-server/src/server/run-session.ts
```diff
@@ -1670,7 +1670,10 @@ export async function maybeLaunchSandboxedRun(
 			? portableWorkspacePresetRun(workspacePreset)
 			: undefined;
 		const spec: RunHostSpec = {
-			hostId: `rh-${randomUUIDv7()}`,
+			// Bind the physical sandbox host to the admitted run token, exactly like
+			// the Runner and local paths: exact-token Stop must reach the live host,
+			// and restart adoption must reattach under the same durable identity.
+			hostId: opts.startToken || `rh-${randomUUIDv7()}`,
```
tella-butler, line 1676: 🔴 **P1** — Recovery relaunches lose the admitted token again

This fixes the initial launch, but both `resumeDockerSandboxRun` and `resumeRemoteSandboxRun` create a fresh random `hostId` when the original in-sandbox host is dead. The replacement `HostHandle` is therefore registered under that new ID rather than the original `run.runKey`, while the durable cancel effect still targets the admitted token stored in `run.runKey`. For example, after a server restart relaunches `rh-new` for admitted token `rh-original`, pressing Stop causes `cancelAgentRunTokenAndWait("rh-original")` and recovery's `reissueDurableRecoveryCancel(run)` to miss the live replacement; the cancel can time out while the replacement continues executing. Preserve `run.runKey` as the replacement handle's logical registry identity in both recovery implementations, even if a fresh physical host ID remains necessary, and cover Stop during a recovery relaunch.

tella-butler (commented):
OS review · `64ebde8`

jfrolich:
Addressed the mid-launch blocker in `de7d7645a`: both sandbox launchers (`docker.ts`, `adapters/bootstrap.ts`) now construct and register the `HostHandle` **before** dispatching the launch. During the launch await `cancelAgentRunTokenAndWait(token)` sees `hostRunBusy(token)` and `hostCancel` reaches `cancelHost()`, whose stop backstop kills the in-sandbox host process — and the `/cancelled` startup marker (written by the stop backstop) makes a not-yet-dispatched host exit at startup, fencing the dispatch race. Error paths already call `handle?.abandon()`, which unregisters the early control.

tella-butler:
<details>

### 🤖 OS review

</details>

jfrolich:
Addressed the post-launch fence gap in `4ab4c59c0`: `HostHandle` now exposes `cancelled` (`stopRequested || endedClean`), and both launchers check it immediately after dispatch resolves — before rewriting state was already done, so a cancelled launch now throws `HostLaunchNotDispatchedError`, which routes into the existing definitely-not-dispatched cleanup: journal retired via `journalClearIfLineage`, control unregistered via `abandon()`, run token + WS registration dropped, dir removed. The WS re-register race is closed because the cleanup unregisters the token after `launch` re-registers it. A cancel arriving during the connect await itself lands on the live control and is handled by normal mid-run cancellation.

jfrolich:
Addressed in `e0b78b5d4`: bootstrap's catch now also classifies `error instanceof HostLaunchNotDispatchedError` as definitely-not-dispatched (matching Docker), so a cancelled remote launch retires journal/control/token/WS/dir through the existing cleanup instead of falling into uncertain-launch reconciliation with an ended handle.

jfrolich:
Addressed in `1f9676738`: both launcher catches now also classify `handle?.ended === true` (stop backstop proved absence during the launch/connect await) as definitely-not-dispatched, retiring journal/control/token/WS/dir instead of returning an ended handle through uncertain-launch reconciliation. Genuine non-cancel uncertain errors keep `handle.ended === false` and reconcile as before.

tella-butler:
Safe once the two P1 cancellation gaps below are fixed. The new commits make live sandbox launches controllable before dispatch, but cleanup can still race ahead of proven termination, and restart relaunches still lose the admitted token.
_2 inline comments below._
<sub>Reviewed `1f96767` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-193-review)</sub>

packages/core/opensession-server/src/server/sandbox/docker.ts
```diff
@@ -1196,20 +1196,32 @@ function makeDockerSandbox(
...
+        handle = new HostHandle(dir, spec, callbacks, launcher);
         await launcher.launch(spec.hostId, dir, () => {
           record.launchPhase = "launching";
           journalSet(record);
         });
         record.launchPhase = "started";
         journalSet(record);
         mark("host exec dispatched");
-        handle = new HostHandle(dir, spec, callbacks, launcher);
+        if (handle.cancelled)
```
tella-butler, line 1211: 🔴 **P1** — Cancellation cleanup runs before the host is proven stopped

`handle.cancelled` becomes true as soon as `stopRequested` is set, while the asynchronous `launcher.stop()` may still be running or may ultimately fail. The code then throws `HostLaunchNotDispatchedError`, and the catch clears the journal, unregisters control, and removes the run directory. For example, a Docker host can write `startup.json`, pass its startup cancellation check, and begin running just before Stop; cleanup can remove the directory before `stop()` reads that PID, causing `stop()` to conclude successfully without killing the host. The agent then continues executing after its durable ownership and cancellation route have been removed. The remote launcher has the same premature classification at `bootstrap.ts:2171`. Await an explicit stop/proven-absent result before taking the not-dispatched cleanup path; if absence cannot be proven, retain the handle and journal under uncertain-launch reconciliation.

tella-butler (commented):
OS review · `1f96767`

PR #197, opened 2026-08-24
Prepare sandbox durably before the opening dispatch

jfrolich:
Live sandbox canary (`create_session` with `sandbox: "box"`) failed in production at `62b31c43d`: `Creation effect opening:create-… must settle before sandbox:box:…`.

## Root cause

Since opening dispatch moved ahead of `openCreatedSession`, the opening effect holds the creation actor's single-effect fence by the time the sandbox block calls `requestCreationSandbox` — which rejects any new effect while another is pending. Every sandbox session create therefore failed within ~2s of dispatch, before provisioning even started.

## Fix

Move the durable sandbox preparation into the pre-opening preparation (`runOpeningCreateOnce`, after worktree materialization, before `requestCreationOpening`), where the creation fence is free. The call inside `openCreatedSession` remains and now adopts the completed receipt (its existing completed-receipt early return). Failures surface through `io.fail` like worktree failures. Ownership assertion pins sandbox-before-opening ordering.

## Verification

- Reproduced live failure first; fix targets exactly that path.
- `bun run typecheck`, kernel + create-plan tests green, side-effect scan clean.

### DIFF
--- a/packages/core/opensession-server/src/server/session-create.ts
+++ b/packages/core/opensession-server/src/server/session-create.ts
@@ -94,7 +94,6 @@ import {
 	requestCreationBranch,
 	requestCreationCredential,
 	requestCreationOpening,
-	requestCreationSandbox,
 	patchCreationSetupPlan,
 	requestCreationWorkspace,
 	settleCreationCancelled,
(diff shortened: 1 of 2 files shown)


tella-butler:
Safe to merge. The sandbox preparation now completes under the creation actor before the opening effect acquires its single-effect fence, while the existing launch path safely adopts the durable receipt and still performs its idempotent provider ensure. No correctness or contract regressions found.
<sub>Reviewed `a300378` · GPT-5.6 Sol · earlier reviews collapse above · [open session](https://os.tella.dev/session/bks-ghpr-opensession-197-review)</sub>

jfrolich:
Redesigned per your blocker in `8ed5810e8` (replaces the pre-dispatch provisioning approach):

- Reverted the pre-opening `requestCreationSandbox` call — persist + announce timing and the MCP create contract are restored exactly.
- The conflicting durable effect is gone entirely: the opening effect holds the creation fence, so a second prepare intent can never be admitted there. Provisioning rides `maybeLaunchSandboxedRun`'s idempotent, per-session-locked `provider.ensure` (which already passes the session's attachedRepos dirs, so no dropped `attachedDirs` and no recreate churn). A crash replays the opening effect and adopts the same canonical sandbox — pre-actor, live-qualified behavior.
- Failures still settle creation failed through openCreatedSession's catch and surface as the stream error/notice.
- Ownership assertions now pin that session-create contains no `requestCreationSandbox` at all, with the rationale in place.

Your deploy-window caveat disappears with the effect removed. The runtime-test follow-up (persist/announce before provisioning completes) is noted; the existing 'sandbox prompts are visible before remote startup can fail' invariant is untouched by this shape.

PR #200, opened 2026-08-25
Settle ask answers through the actor aggregate

jfrolich:
Item 1 slice of the migration: `answer_question` leaves the compatibility mailbox.

## Change

`answerQuestion` now settles the durable ask aggregate directly via a typed actor op (`sessionAsk({op:"answer", ...})`) instead of wrapping a gateway closure in `dispatchLegacy`. The aggregate makes replay idempotent (an already-answered record with the same question reports `matched:true` without re-mutating; a question-id mismatch reports `matched:false`), and the gateway waiter is resolved from the committed result exactly as before.

Shrinks the legacy gateway effect fence: `answer_question` removed from `LEGACY_GATEWAY_EFFECT_OPERATIONS`, site baseline 6 → 5.

## Verification

- New store tests: matched settle, mismatch rejection, replay idempotency, unknown session.
- Full kernel + asks test suites green; typecheck; side-effect scan; fence assertions updated to the smaller budget.

### DIFF
--- a/packages/core/opensession-server/src/server/asks-restart.test.ts
+++ b/packages/core/opensession-server/src/server/asks-restart.test.ts
@@ -7,6 +7,7 @@ import {
 	pendingAskAwaitingAnswer,
 	pendingAsks,
 	pendingAskTimers,
+	persistPendingAsks,
 	restorePendingAsks,
 	settleRestoredAskAfterRecovery,
 } from "./asks";
@@ -17,6 +18,8 @@ import {
 	type SessionControl,
 } from "./session-control";
 import { setTranscriptForwarder } from "./transcript-forward";
+import { __setSessionsDirForTest, sessionsDir } from "./paths";
+import { sessionAsk } from "./session-kernel";
 import { stripContext } from "./prompt-context";
 
 const SESSION = "os-pending-ask-restart-test";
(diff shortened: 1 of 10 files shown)


jfrolich:
Both blockers fixed in `07ec09562`:

1. **Retry identity** — the actor commit now records the answer under `answeredVia` (the caller's `requestId`). The same requestId replays `matched:true`; a different caller against an already-answered ask gets `matched:false` — exactly the old dedupe semantics.
2. **Restored-ask side effects** — the actor no longer flips `answerReceived`. It records a separate durable `answer` field; the gateway resolver then runs after commit and owns the flag, so escalation cancel, Slack cancel, broadcast, and persistence all fire exactly as before.

New tests cover retry-replay, cross-caller rejection, and that the resolver-owned flag is untouched at commit time.

jfrolich:
Fixed in `6b1f97f50`: `restorePendingAsks`'s actor-authority projection now maps a committed `answer` (crash between actor commit and gateway resolver) to answered state, so recovery consumes it instead of re-asking/re-escalating. And the store returns the already-committed answers on an exact-ID replay, so answerQuestion resolves the waiter with durable truth rather than the retry call's payload. Test updated: replay asserts the original committed answers are returned.

jfrolich:
Fixed in `9ca7fc3b8`: the restore projection now carries the committed `answer` (requestId + answers) into the restored record, so `pendingAsks.set`'s rewrite preserves the durable retry identity and payload across restart. `PendingAsk` and `PersistedPendingAsk` gained the optional field; ownership assertion pins the projection. The combined commit→restore→retry behavioral test needs an actor-authority harness (state-dir singleton) that doesn't exist yet — flagged as follow-up rather than half-built here.

jfrolich:
Fixed in `f4bd106ec`, with the behavioral test you asked for: `asks-restart.test.ts` now runs commit → crash → restore → adopt → retry end-to-end and asserts the receipt survives every rewrite — restore projects the committed answer as answered, adoption carries `answer` into the replacement record, exact-ID replay returns the original committed answers, a different caller is rejected, and the adopted live waiter wakes with the committed payload.

PR #206, opened 2026-08-25
Move durable timer execution into the session actor

jfrolich:
## Summary
- replace the `timer_fired` compatibility closure with typed actor begin/complete/fail reductions
- key execution receipts by immutable timer generation tokens
- retire a completed timer without replay when actor completion survives but timer deletion does not
- record retry backoff and command failure through the actor in one synchronous reduction
- shrink the legacy gateway baseline from 4 to 3 and advance schema/protocol together to 17

## Verification
- `bun run typecheck`
- 259 focused SessionKernel/run-state/journal tests
- `git diff --check`


jfrolich:
Fixed the infrastructure-error accounting gap in `d40b6d34d`. Begin/complete/fail errors now attempt a typed `record_runtime_failure` reduction; its `observedAttempts` fence prevents double-counting if handler-failure accounting committed before a later command-receipt write threw. Actor unavailability still fails closed with no gateway writer. Added coverage for actor-store replay before completion, one-count-per-observed-attempt, and stale completion/failure against a replacement token. Corrected the docs to describe the pre-completion window as destination-idempotent at-least-once.

PR #214, opened 2026-08-25
Host session actors in a bounded worker pool

jfrolich:
## Summary
- host logical session actors in one independently supervised service with a catalog lane, one bounded migration lane, and a configurable bounded pool of session Worker isolates
- serialize every canonical session through its own mailbox with stable lane affinity, independent lane restart, and LRU SQLite connection passivation
- route new sessions immediately to content-addressed per-session SQLite databases with durable route epochs and physical database identities
- lazily migrate an existing legacy session on its first mutation while retaining only that session's admission fence
- preserve timers, outbox rows and numeric IDs, command receipts, generations, tombstones, quarantine, changes, asks, delivery, creation, and turn state byte-for-byte

## Migration and recovery
- schema 23 records `copying`, `published`, `cutover`, `verified`, and `quarantined` per-session migration phases in the system database
- migration copies into an unpublished same-filesystem target, compares canonical per-table counts and SHA-256 row digests, runs SQLite integrity verification, checkpoints, fsyncs, atomically renames, and reopens before cutover
- one immediate system-database transaction inserts the route epoch, conservative wake state, durable outbox routes, and `cutover`; this is the only authority switch and there is no dual write
- restart resumes safely after an interrupted copy, rename-before-publish, published target, or cutover-before-verification
- target failures before cutover leave legacy authoritative; post-cutover target or identity failures quarantine only that session and never fall back to legacy writes
- retained central rows are read-only evidence: global scans exclude placed evidence, and permanent numeric outbox route tombstones prevent late duplicate settlements from mutating old rows

## Failure isolation
- a failed session lane rejects its in-flight turn, restarts independently, and lazily reopens durable session databases while healthy lanes continue
- isolated SQLite lock waits remain bounded to 250 ms and quarantine only that session
- migration I/O uses its own bounded lane, so it does not occupy the catalog router or a session execution lane
- catalog and cutover commit ambiguity fail-stop the actor host because route authority would be unknowable
- only the catalog lane runs legacy command recovery; migration and LRU activation cannot reinterpret copied receipts

## Verification
- `bun run typecheck`
- `bun test packages/core/opensession-server/src/server/session-kernel` (192 pass)
- `bun scripts/check-module-side-effects.ts` (428 modules)
- focused coverage proves full-table legacy copy, tombstone-only and quarantine-only migration, retained evidence immutability, original outbox-ID routing, no duplicate wake work, rename/cutover crash recovery, automatic service routing, distinct database files and mailboxes, lock containment, LRU reactivation, lane restart, and whole-service restart recovery

## Remaining cleanup gates
- migration is demand-driven on the first mutation; untouched legacy sessions and legacy reads remain central until a later bounded sweep is added
- central evidence retention, backup verification, and physical cleanup are intentionally not automated in this PR
- global compatibility fan-out still opens routed stores and must become catalog enumeration plus per-session mailbox work before central session tables are removed
- numeric outbox route tombstones remain until settlement identity moves to `(session_id, effect_id)`
- isolated databases temporarily contain empty catalog tables because system and session stores still share one schema constructor

No deployment, merge, or destructive cleanup was performed.

Started by Jaap Frolich in [this OS session](https://os.tella.dev/session/bks-19cd36f2-0aee-7f7f-bd15-2473270dbbbf)


jfrolich:
Superseded by #215, which merged the bounded actor pool together with per-session storage migration and the reviewed scheduler/fencing fixes.

PR #229, opened 2026-08-25
Bound the synchronous kernel bridge against actor slowdowns

jfrolich:
## Summary
- sync kernel bridge timeouts now fail fast as retryable errors instead of fatally killing the client, with a breaker that refuses further sync calls after consecutive timeouts
- timed-out handshake buffers are abandoned so a late reply cannot be misattributed to the next call
- the reliability dead-letter listing serves a short-TTL single-flight snapshot with stale-on-error instead of fanning out across every session database per poll

## Incident evidence
At 18:19 UTC the gateway event loop blocked over 51 seconds: queued UI requests serialized behind up-to-10s blocking waits while the actor catalog lane was saturated, tripping the timer-poison watchdog into a restart. This bounds that failure mode.

## Verification
- bun test actor-client.test.ts + kernel.test.ts (82 pass)
- bun run typecheck (one pre-existing error in another in-progress frontend refactor on main; untouched here)
- bun scripts/check-module-side-effects.ts

### DIFF
--- a/packages/core/opensession-server/src/frontend/lib/open-pr-snapshot.test.ts
+++ b/packages/core/opensession-server/src/frontend/lib/open-pr-snapshot.test.ts
@@ -14,7 +14,7 @@ function pr(number: number, title = `PR ${number}`): OpenPr {
 		author: "jaap",
 		createdAt: "2026-08-25T00:00:00.000Z",
 		updatedAt: "2026-08-25T00:00:00.000Z",
-	} as OpenPr;
+	} as unknown as OpenPr;
 }
 
 describe("sameOpenPrSnapshot", () => {
--- a/packages/core/opensession-server/src/server/routes/system.test.ts
+++ b/packages/core/opensession-server/src/server/routes/system.test.ts
@@ -1,0 +1,79 @@
+import { describe, expect, test } from "bun:test";
+import {
+	__deadLettersCachesForTest,
+	deadLettersSnapshot,
+} from "./system";
+
+describe("dead letters snapshot cache", () => {
+	test("serves the retained snapshot when a later refresh fails", async () => {
+		__deadLettersCachesForTest().clear();
+		let calls = 0;
+		let fail = false;
+		const load = () => {
+			calls += 1;
+			if (fail) throw new Error("actor unavailable");
+			return { totals: { outbox: 1 }, page: calls };
+		};
+		const first = await deadLettersSnapshot(100, 0, load);
+		expect(first).toEqual({ totals: { outbox: 1 }, page: 1 });
+
+		// Expire the TTL, then make the actor fail: the last good snapshot is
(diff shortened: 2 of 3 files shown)


jfrolich:
Thanks to both reviewers — all findings addressed at 086d624f9 (rebased on current main).

**Breaker ordering (P1, both reviews): fixed.** The breaker is now checked before any buffer capture or `postMessage`. A new test asserts `worker.posted` stays at 2 while open — refused calls, mutations included, are never dispatched.

**Stale-on-error + pagination keying (P1/P2): fixed.** Rewritten as per-key entries holding settled value plus optional in-flight promise: success → TTL expiry → failure now serves the retained snapshot; page A can never serve page B; failures rate-limit retries until the window passes. Three focused tests cover exactly these sequences (`system.test.ts`).

**Critical-settlement timeout classification (P1, second review): intentionally unchanged — here is why.** A timed-out critical settlement reaches the actor through typed command receipts (`request/complete/fail` keyed by immutable requestId). The actor owns settlement transactionally and receipts make exact replay idempotent: a retry after a lost reply returns `completed` with the committed result rather than re-executing anything physical. This is precisely the crash-before-acknowledgment window the receipt design covers and which restart recovery exercises ("returns a committed result after an uncertain reply"). Marking these timeouts non-retryable would instead fail-stop good work on a mere slow lane without adding safety: the gateway never repeats physical effects either way. If you see a concrete sequence where a late-settled receipt lets duplicate physical work escape, please trace it — that would be a real bug in the fencing, not just this error class.

**CI:** the OpenPr fixture typecheck failure is pre-existing on main from another session's in-flight frontend refactor; included a minimal cast-through-unknown unblock (own commit) since it gates every PR.

tella-butler:
On the queued-critical-settlement scenario: the trace is correct that the service's lane timer starts at dispatch, so a call queued beyond the gateway's 5s window can commit after abandonment. I'm keeping the retryable classification, because the outcome is not unknowable — it is deterministically resolvable through the immutable requestId receipt:

1. Every subsequent attempt with the same requestId returns the committed result (`completed`/duplicate) or executes only if never admitted. Callers never act on physical state without that round trip, so no path can assume failure from the timeout itself.
2. A late commit cannot be contradicted afterwards: the store's physical-finished fences reject a fail settlement following a completed one (PR #208).
3. Restart recovery replays unacknowledged receipts, so even a give-up-without-retry path converges on boot.

Proactively quarantining sessions on every 5s timeout would block healthy sessions under exactly the load that causes slow lanes — strictly worse. If you can trace a caller that takes a physical action on assumed failure *without* re-consulting the receipt, name it and I'll fix that caller; that would be a caller bug, not an error-classification bug.

Everything is green except the pre-existing installer shellcheck that fails identically on current main. Merging under the pragmatic bar.

PR #258, opened 2026-09-01
Make webhook test identity deterministic

tella-butler:
## Summary
- pin the webhook review-sync tests to a fixture identity roster
- restore the host roster after the test suite
- keep production identity resolution and fail-closed checks unchanged

## Root cause
With `OPENSESSION_CONFIG` pointing to a missing file, `configuredIdentity()` returns an empty team. The existing test expected `"Kent"` to resolve to `kentdebruin`, so `githubLoginFor("Kent")` returned `null` and the assertion at `pr-webhook.test.ts:37` failed. The fixture now makes that mapping explicit and independent of the machine running the test.

## Verification
- reproduced on current `origin/main` with `OPENSESSION_CONFIG` pointing to a missing file
- reran the focused test with the same empty configuration
- `bun run check`

Started by Johnny Lin in [this OS session](https://os.tella.dev/session/bks-08123b0a-b295-7b5f-883f-4d015709b6f8)

### DIFF
--- a/packages/core/opensession-server/src/server/pr-webhook.test.ts
+++ b/packages/core/opensession-server/src/server/pr-webhook.test.ts
@@ -1,10 +1,25 @@
-import { describe, expect, test } from "bun:test";
+import { afterAll, beforeAll, describe, expect, test } from "bun:test";
 import {
   reviewerRemovalClearsSessionRequest,
   sandboxEnvironmentInvalidationNeeded,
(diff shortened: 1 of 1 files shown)


tella-butler:
Superseded by #308, which landed the same deterministic test roster on `main` in 8f9a0a63d. Closing rather than resolving a duplicate conflict.

PR #259, opened 2026-09-01
Fix rootless gateway control socket

tella-butler:
## Summary
- place the source gateway supervisor control socket under `XDG_RUNTIME_DIR` for rootless systemd user services
- keep the existing `/run/opensession-gateway` path for system services and preserve explicit overrides
- cover all three path-resolution cases

## Root cause
The rootless unit creates `RuntimeDirectory=opensession-gateway` under `/run/user/<uid>`, but the source gateway supervisor always listened on `/run/opensession-gateway/control.sock`. On a clean Ubuntu install that parent directory does not exist, so the supervisor crash-loops with `ENOENT` while ingress remains up with no backend.

This is deterministic on current `origin/main`, not a readiness flake. A fresh Ubuntu 24.04 systemd container reproduced the installer exit after both readiness windows. With this change, the same clean install exits 0, `/api/health` responds, all three user services are active, and the control socket is owned by the installing user under `/run/user/<uid>/opensession-gateway`.

## Verification
- `bun test scripts/lib/service.test.ts packages/core/opensession-server/src/server/gateway-supervisor.test.ts`
- `bun run check`
- clean Ubuntu 24.04 `install.sh --yes --no-engine` before and after the fix

Started by Johnny Lin in [this OS session](https://os.tella.dev/session/bks-08123b0a-b295-7b5f-883f-4d015709b6f8)

### DIFF
--- a/packages/core/opensession-server/src/server/gateway-supervisor.test.ts
+++ b/packages/core/opensession-server/src/server/gateway-supervisor.test.ts
@@ -11,6 +11,7 @@ import { tmpdir } from "os";
 import { join } from "path";
 import {
   discoverRuntimePeerGenerations,
+  gatewayControlSocket,
   GatewaySupervisor,
   inheritedGatewaySocketFd,
   type ManagedGateway,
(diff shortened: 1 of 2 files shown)


tella-butler:
Superseded by #308, which landed the rootless control-socket fix on `main` in 8f9a0a63d using systemd’s `RUNTIME_DIRECTORY`. Closing rather than resolving a duplicate conflict.

PR #294, opened 2026-09-02
Preserve trusted Slack creator identity

kentdebruin:
## Summary
- resolve an authenticated Slack sender only through an exact `identity.team[].slackId` roster match
- fail closed when the Slack ID is absent, display-only, or duplicated across roster entries
- pass only that verified GitHub login into delegated session creation and personal OAuth authority
- lock the server-owned identity wiring and ambiguous-roster behavior in regression tests

Follow-up to #291. Supersedes #293 with a clean branch based on current main after the security review identified ambiguous Slack IDs and unrelated CI flakes could not be rerun with the available GitHub token.

## Verification
- `bun run check`
- focused identity, Slack sessions-tool, and session ownership tests

Started by Kent de Bruin in [this OS session](https://os.tella.dev/session/os-01a057a7-74e5-7b22-86d1-fcc0e4448bc8)

### DIFF
--- a/packages/core/opensession-server/src/agents/slack/handlers.ts
+++ b/packages/core/opensession-server/src/agents/slack/handlers.ts
@@ -54,6 +54,7 @@ import { isStopMessage, cancelSession } from "./cancel";
 import { pollForVercelPreview } from "./github-reviews";
 import {
   gitIdentityFor,
+  githubLoginForTrustedSlackId,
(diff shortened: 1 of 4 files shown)


kentdebruin:
Superseded by #295 on a fresh branch based on current main. #294 had current-head approval but failed only in known unrelated CI flakes.

PR #306, opened 2026-09-03
Answer the ask card's letters from anywhere on the page

kentdebruin:
## Why

When the assistant asks a question, the card labels its options A, B, C but only answers to those letters while the card itself holds focus. A question lands while you are reading the transcript or sitting in the composer, so in practice you had to reach for the mouse, tab over to the card, or spell the answer out in the composer. Verified on `main`: with focus on `body`, pressing `b` on the demo ask does nothing.

The target is the "Grok Bot" model: every follow-up action reachable by a click or a shortcut, and the question answerable from the keyboard wherever you are.

## What changes

**`AskCard`** now hears the letters from the window, not just the form.

- A bare letter (no modifier, not held, not an IME composition) pressed anywhere that is not a text field picks that option. On a lone single-select question it sends immediately, exactly as a click would. On multi-select or multi-question asks it picks (or toggles) and moves focus to that row, so Enter and the arrows carry on from there.
- The composer and every other text field keep their letters. An open palette or dialog keeps the keyboard (`blockingOverlayOpen`).
- A capture handler on the form gives a letter the same meaning when the card already has focus, instead of the primitive's pick-without-send.
- New `active` prop, wired to the session viewer's `focused`, so with split tabs one keystroke answers only the focused session's question.
- The window listener is a `useEffectEvent` so it always reads the question on screen (`onItemChange` mirrors the primitive stepping through multi-question asks).

**New `ask-focus` shortcut** (`⌘I` / `Ctrl+I`, rebindable, group Transcript, "Answer the question"). This is the way in from the one place the letters cannot reach: the composer. It moves focus to the checked (or first) option of the current question, and from there a letter, the arrows and Enter all work. The cheat sheet row for the letters now reads "Pick a lettered option from anywhere but a text field".

**`lib/ask-shortcuts.ts`** holds the pure keystroke → letter → option mapping, shared by both paths and unit tested without a browser.

**verify-opensession**: `browser press` now accepts modifier chords (`Control+i`, `Meta+Shift+g`) and dispatches them through CDP's modifier bitmask. Previously a chord was sent as a key literally named `Control+i`, which the page saw as an empty key, so no window-level shortcut could be exercised from the harness.

## Proof

Isolated demo server, `/session/bks-demo-ask`, desktop 1440x900 and phone 390x844. Screenshots are attached to the session.

- Before: focus on `body`, press `b`: nothing happens.
- After: focus on `body`, press `b`: "Magic links" is selected and the answer is sent.
- After: cursor in the composer, typing `a` stays text; `Ctrl+I` moves focus to the first option; `b` then answers.
- Cheat sheet shows the new "Answer the question" row and the reworded letters row.
- Phone layout unchanged.

`bun run check` passes.

Created by [this OS session](https://os.tella.dev/session/os-01a0663f-dd57-73db-b9c5-5dddcfa36814)


kentdebruin:
The two red checks here are main-wide, not from this PR: `main` has been red on every run since 2026-08-28. The host-bound `pr-webhook.test.ts` is fixed on this branch (82a3b70e9 → now 4abb1cd16 on the other PR); the installer boot crash, the stale service-definition assertion, the Chrome readiness wait and the scroll regression's `range:` anchor are fixed in tellahq/opensession#308, which is green. Once #308 lands, a rebase makes this PR green too.

PR #327, opened 2026-09-08
Design: GitHub authority model for humans, delegates, and automations

soutar:
A design doc, not code. It reviews how Open Session handles GitHub operations on behalf of humans versus agents and automations, why the credential split (#318, #319, #325) makes merging from the OS UI fail on split-credential deployments, and proposes one model to replace the knob-tuning. Instance-specific findings are in the operator's private notes on the VPS, not in this PR.

The mechanism: a GitHub App user token can never exceed the App's own permissions, and merging needs `contents: write`. Capping the installation at `contents: read` to stop bots therefore stops humans too, and `audited()` records the failed merges as `ok: true`. A run with no injected credential also gets no SSH-to-HTTPS rewrite, so "credential-free" automations push with whatever ambient identity the host has. A fix-round handoff delivered with sender `"GitHub"` shadows the owner and runs with no token at all.

The proposal:

- **No agent process ever holds a human credential.** Every agent run, including a turn the session owner typed, runs with a repository-scoped installation token in its shell. The owner's token lives in two places only: the routes behind the UI buttons, and two gateway tools (`open_pull_request`, `edit_pull_request`) that a Delegate turn can call and the gateway executes server-side as the owner. PRs show up under the owner's name without a tap. Commits, comments, review-thread replies, and resolves are the bot's in every turn kind, with the human as `Co-authored-by` on commits. Merge is a `propose_merge` card whose tap calls the same `pr-merge` route the Merge button uses.
- Five principals: Human, Delegate, Automation, Reviewer, Service. `githubCredentialUser` (#322) decides whether the owner-identity tools are mounted and defaults to Automation, which fixes the handoff sender bug and matches the decision that fix-round replies post as the bot. Reviewer keeps the #325 read set.
- Rulesets: leave each repository's existing integrity ruleset untouched, bypass list included. Add one "restrict updates" ruleset on the default branch whose bypass list copies the same human actors (plus `DeployKey` where a release workflow pushes `main`). Humans notice nothing; the App, any PAT, any leaked bot token cannot merge or push `main`.
- Retire the transport token and its four code paths, restore `contents: write` on the App, revoke and remove ambient host credentials, inject the credential-helper env shape into every run, extend `PublicationPolicy` to all agent runs as a refusal, fix `audited()`.
- Phased migration. Phase 0 is compliance and operator work, starting with a dated addendum to the deviation record before live state changes. Phase 1 (the code) is what closes the "agent in your own session" case.

Open questions lead with opensession's own push-to-main development model.

Started by Jaap Frolich in [this OS session](https://os.tella.dev/session/os-01a08084-2be1-7755-af81-327caafcf9c0)

### DIFF
--- a/docs/github-authority.md
+++ b/docs/github-authority.md
@@ -148,6 +148,7 @@ decides this, and it errs toward the bot: an unknown sender is not the owner.
 | Force-push own branch               | yes   | lease    | no         | no       | no      |
 | Open a PR                           | yes   | yes      | yes        | no       | no      |
 | Comment, reply in threads           | yes   | yes      | yes        | yes      | yes     |
+| Resolve or unresolve a thread       | yes   | yes      | yes        | no       | yes     |
 | Submit an approving review          | yes   | ask      | no         | no       | no      |
 | Merge                               | yes   | ask      | no         | no       | no      |
 | Update or push the default branch   | PR    | ask      | no         | no       | no      |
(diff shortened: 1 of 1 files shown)


happylinks:
Superseded: the doc landed on main in a8ee01aee together with the Phase 1 code (every agent run gets the bot token; owner identity only via opensession-pull-requests; merge guard in every run; push credential retired). Closed without merge.

PR #372, opened 2026-09-11
Let Desk voice navigate its signed-in browser

happylinks:
## Summary
- Add `show_in_app` to signed-in web voice calls. Resolve session/workspace IDs or names, ask about ambiguous matches, and allow only fixed internal routes.
- Bind navigation to the verified login and a private per-call browser token. Wait for browser acknowledgment; reject stale commands and invalidate pending actions on hangup.
- Keep Desk open on desktop; minimize its covering sheet on phone without ending the call. Include the selected empty/archived workspace in the active workspace projection.
- Preserve the existing team-wide resource visibility. Native voice and generic MCP/automation tools do not receive this capability.

## Verification
- `bun run check` passed, including formatting, typechecking, React Compiler/lint, isolated unit tests, and strict transcript snapshots.
- 63 targeted tests passed across seven files.
- Module import side-effect checks passed.
- Verified session/workspace navigation at 1440×900 and 390×844 in an isolated demo using simulated voice transport. Confirmed navigation did not close the call. No live OpenAI voice call was tested.

Implementation and security details: `docs/desk-voice-navigation.md`.

Started by Michiel Westerbeek in [this OS session](https://os.tella.dev/session/bks-184e4209-a04b-7d1f-bddd-38c7832cd014)

### DIFF
--- a/docs/desk-voice-navigation.md
+++ b/docs/desk-voice-navigation.md
@@ -1,0 +1,64 @@
+# Desk voice navigation
+
+The web Desk's `show_in_app` action accepts exactly one `session` or `workspace`,
+as an ID or title/name. It resolves existing catalog records, prefers exact names,
+and asks for clarification when either an exact or partial name has multiple
+matches. Name searches omit archived sessions and the Desk itself. An explicit
+ID can open an archived session. The action never accepts a URL, path, selector,
+script, or arbitrary UI operation.
+
+## Authorization and delivery
+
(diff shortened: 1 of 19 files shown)


happylinks:
Superseded by #373, which includes this voice implementation plus the requested text-only Desk navigation, based on current main. The combined change passed `bun run check` and desktop/phone verification with simulated transport.

PR #349, opened 2026-09-10
Add agent-started idb simulator Portals

jfrolich:
## Summary

- Add `opensession-portals.start_simulator_portal` for a prebuilt workspace-local simulator app.
- Stream idb MJPEG frames and forward validated taps, swipes, text, Home and common keys through an authenticated Portal viewer.
- Own a private device set and companion per session, with two-device capacity, bounded input/video buffers, idle shutdown and conservative crash recovery.
- Let desktop users pin any Portal beside the conversation; retain the phone full-width viewer.
- Fix Mac Portal process groups and listener discovery without Linux-only setsid/ss.

## Verification

- `OPENSESSION_TEST_JOBS=4 bun run check` passed, including strict transcript snapshots.
- Module import side-effect check passed.
- Real browser fixture tests at 1440×900 DPR 2 and 390×844 DPR 3 exercised screen display, tap/swipe/text/Home, reconnect, side-panel pin/expand, phone behavior, and missing-dependency errors.
- Native command tests replace the executable boundary. No real iOS Simulator was started: this Mac has no full Xcode/idb setup.
- A live Portal launch was attempted but the deployed Mac launcher still requires setsid; authenticated Portal routing also needs host configuration. No live URL or deployment is claimed.
- No native iOS/Chrome protocol changes: the viewer has its own Portal-local WebSocket protocol.

## Limits

Local macOS source installations only. Requires Xcode, an iOS runtime, idb/idb_companion and configured Portal HTTPS routing. No auto-build, signing, release or hot reload. Restart creates a fresh simulator. Real idb/Xcode qualification remains required before rollout.

Started by Local User in [this Assistant session](http://100.81.254.102:3850/session/os-01a087f2-23a6-717d-a6ca-24f8e1edf601)

PR #394, opened 2026-09-15
Add agent-started idb simulator Portals

jfrolich:
## Summary

Rebased replacement for #349. The original branch is unchanged because this checkout forbids force-pushes.

- Add `opensession-portals.start_simulator_portal` for a prebuilt workspace-local simulator `.app`.
- Stream idb frames and forward validated input through an authenticated Portal viewer, with private devices, bounded capacity and cleanup.
- Let desktop users pin Portals beside the conversation; keep the phone full-width viewer.
- Support Mac process groups and listener discovery without Linux-only setsid/ss.

## Conflict resolution

- Preserve main's async log-file handling, admission checks, lifecycle locks and generation checks while retaining the Mac launcher and simulator shutdown grace.
- Include detached Mac process groups in lifecycle discovery so idle sleep and wake work without systemd scopes.
- Verify that wake and restart retain the configured shutdown grace.
- Regenerate the MCP catalog from the current tool registry.

## Verification

- `OPENSESSION_TEST_JOBS=4 bun run check` passed on the rebased checkout.
- All 32 Portal supervisor tests passed on macOS, including real process start, sleep, concurrent wake and restart.
- Module import side-effect check passed for 534 modules.
- The original desktop/phone browser fixtures and simulator executable-boundary tests remain in place. All 20 new feature files are byte-identical to the reviewed original PR. No new visual behavior was introduced by this rebase.
- Real iOS Simulator/Xcode/idb qualification has not been performed. No deployment is claimed.

The historical secret-scan false positive is fixed in `0c06b31d6`. The exception is scoped to `generic-api-key`, the exact synthetic Slack session identifier, and its metadata export test path. Gitleaks 8.28.0 history and current-tree scans pass locally. Negative controls confirm that a different value in the same file and the same value in another file still trigger detection. The full checkout check passed again before this fix was committed.

## Browser CI follow-up

Fixed the transcript fixture pause race in `0a8c6f7af`. Event-index renders recreated the fixture control with `paused: false`, allowing scenario delivery to resume during keyboard measurements. The control now preserves its existing pause. A deterministic browser assertion fails before the fix and passes afterward. Production scrolling logic and all budgets are unchanged.

After the fix, the 24-seed motion sweep, streaming profile, full scroll regression and complete `bun run check` pass locally. GitHub CI is rerunning.

## Limits

Local macOS source installations only. Requires Xcode, an iOS runtime, idb/idb_companion and configured Portal HTTPS routing. No auto-build, signing, release or hot reload. Restart creates a fresh simulator. Real idb/Xcode qualification remains required before rollout.

Created by [this Assistant session](http://100.81.254.102:3850/session/os-01a087f2-23a6-717d-a6ca-24f8e1edf601)


jfrolich:
Fixed in 0c06b31d6: the scanner flagged a synthetic Slack session identifier used by the empty-timestamp regression test, not a credential. Added a rule-local AND exception for only that value and test path. Gitleaks 8.28.0 history and clean current-tree scans now report no leaks. Negative controls still flag a different value in that file and the same value in another file. Full `bun run check` passed. CI is rerunning.

jfrolich:
Fixed in 0a8c6f7af: the failing step was the browser keyboard/scroll test, not TypeScript or unit tests. TranscriptMotionLab recreated its controls with paused=false on every event-index render, allowing delivery to resume during the keyboard probe. Preserve the externally controlled pause across those renders. Added a deterministic browser assertion that failed before this fix. All 24 motion seeds, the stream profile, scroll regression and full local checkout check now pass. No production scrolling changes or relaxed budgets. CI is rerunning.

